PR #43, opened 2026-04-04
feat(navdata): add Signal K Discovery protocol and WebSocket support

### LINKED ISSUE #42: Support Signal K Discovery protocol and WebSocket connections
## Summary

Add support for the Signal K [Discovery and Connection Establishment](https://signalk.org/specification/1.7.0/doc/connection.html) protocol, and WebSocket (WS/WSS) as a transport for receiving navigation data.

## Motivation

Currently Mayara connects to Signal K servers only via plain TCP (`_signalk-tcp._tcp` mDNS). This has two limitations:

1. **No version negotiation** — mDNS discovery connects directly without querying which API versions the server supports. The Discovery protocol (`GET /signalk`) returns versioned endpoints, allowing the client to select the correct one.

2. **No WebSocket/TLS support** — Some Signal K servers (only expose HTTPS/WSS endpoints and do not advertise `_signalk-tcp._tcp` at all. These servers are invisible to Mayara, preventing heading, position, and AIS data from flowing.

## Proposed changes

- Browse mDNS for `_signalk-https._tcp` and `_signalk-http._tcp` services
- Use the REST discovery endpoint (`GET /signalk`) to obtain available transports (TCP, WS, WSS)
- Prefer TCP for backward compatibility, fall back to WebSocket
- Support `--navigation-address ws:<addr>:<port>` and `wss:<addr>:<port>` for explicit connections
- Add `--accept-invalid-certs` flag for self-signed certificates
- New `signalk` module for discovery protocol, WebSocket connection, and TLS handling

msallin:
> **Draft PR** — wait-for-feedback state. All 4 blockers from @dirkwa's review have been addressed (see latest push). Ready for another pass; I'll keep it as draft until the approach is confirmed.

## Summary

- Implement Signal K [Discovery and Connection Establishment](https://signalk.org/specification/1.7.0/doc/connection.html) protocol
- Add WebSocket (WS/WSS) transport for receiving navigation data from Signal K servers
- New `signalk` module (`src/lib/signalk.rs`) isolates discovery, HTTP client, TLS, and WebSocket handling
- Browse mDNS for `_signalk-tcp._tcp`, `_signalk-http._tcp`, and `_signalk-https._tcp` concurrently — backward compatible with existing anonymous TCP installs

Closes #42

## Motivation

Modern Signal K servers (e.g. OpenPlotter) often only expose HTTPS/WSS endpoints and do not advertise `_signalk-tcp._tcp`. Mayara cannot currently discover or connect to these servers, preventing heading, position, and AIS data from flowing and breaking true-motion trails, ARPA tracking, and CPA computation.

## Changes

- **mDNS (additive):** Three concurrent service browses; the `_signalk-tcp._tcp` path is preserved so anonymous TCP-only installs keep working
- **Discovery:** `GET /signalk` over a hand-rolled HTTP/1.1 client built on `tokio` + `tokio-rustls` (no `reqwest` runtime dep)
- **Transport preference:** WS > TCP fallback — WebSocket is the only Signal K transport that can carry authentication in the future
- **TLS:** `wss://` uses a permissive certificate verifier gated behind `--accept-invalid-certs`, appropriate for boat-LAN self-signed setups. `rustls-native-certs` dropped entirely
- **CLI:** `--navigation-address ws:/wss:` and `--accept-invalid-certs`, plus updated help text documenting the full `tcp:/udp:/ws:/wss:` matrix
- **401 hint:** A friendly log line pointing at #42 when a WS upgrade is refused with HTTP 401
- **Docs:** `USAGE.md` updated with the new transport matrix and authentication note

## Summary

This pull request adds Signal K Discovery protocol support and WebSocket (WS/WSS) transport for navigation data connections, addressing the need for authenticated navigation server integration alongside existing TCP/UDP support.

## Key Changes

**New Signal K Module** (`src/lib/signalk.rs`)
- Implements Signal K Discovery and Connection Establishment flow per specification
- Performs concurrent mDNS browsing for `_signalk-tcp._tcp`, `_signalk-http._tcp`, and `_signalk-https._tcp` services
- Issues hand-rolled HTTP/1.1 GET /signalk requests via tokio and tokio-rustls to discover versioned endpoints
- Parses discovery responses to extract advertised WebSocket and TCP endpoints, preferring v1 formats
- Provides permissive TLS certificate verification gated behind `--accept-invalid-certs` flag for self-signed boat-LAN certificates
- Maintains a cached rustls::ClientConfig via OnceLock for efficient connection reuse
- Bounds discovery HTTP reads to 64 KiB maximum response size
- Includes comprehensive unit tests for discovery parsing, transport preference, and authentication observability

**Navigation Data Enhancements** (`src/lib/navdata.rs`)
- Extends navigation data handling to support WebSocket streams alongside existing TCP/UDP
- Routes `ws://` and `wss://` addresses to WebSocket connection logic
- Implements WebSocket-first transport preference with TCP fallback for legacy installations
- Centralizes subscription logic into `signalk_actions_for_line`, shared between TCP and WebSocket transports
- Refactors Signal K delta parsing to iterate through all update and values entries instead of only the first
- Resets cached own-ship context on session reconnection to prevent AIS data misrouting
- Expands logging for connection endpoints

**CLI and Documentation Updates** (`src/lib/mod.rs`, `USAGE.md`)
- Adds new `--accept-invalid-certs` boolean flag to allow invalid TLS certificates
- Expands `--navigation-address` documentation to cover `tcp:`, `udp:`, `ws:`, and `wss:` transport schemes
- Clarifies authentication behavior: anonymous-only for TCP, authentication-capable for WebSocket
- Documents TLS SNI handling limitations for WSS connections
- Includes notes on access-request workflows for authenticated servers

**Error Handling** (`src/lib/radar/mod.rs`)
- Introduces `RadarError::SignalK(String)` variant for Signal K-specific errors

**Test Updates**
- Updates all replay integration tests to explicitly set `accept_invalid_certs: false` in test configuration, ensuring consistent test behavior across the suite

## Transport Behavior

- **Discovery Priority**: WebSocket (WS/WSS) attempted first when advertised; falls back to TCP if unavailable
- **mDNS Compatibility**: Preserves existing `_signalk-tcp._tcp` browsing while adding HTTPS/HTTP variants, maintaining backward compatibility with anonymous TCP installations
- **Authentication**: WebSocket transport enables authentication via access-request flows, while TCP remains anonymous-only
- **TLS Security**: WSS connections use a permissive verifier when `--accept-invalid-certs` is enabled, supporting development and boat-LAN scenarios

## Technical Notes

- Removes reqwest as a runtime dependency; uses tokio and tokio-rustls for HTTP discovery and TLS operations
- Implements hand-rolled HTTP/1.1 client for discovery to maintain spec compliance and support non-default endpoint paths
- Provides friendly INFO log hints on WebSocket 401 responses, directing users toward authentication workflows
- Plans for expanded auth support (persistent clientId, token caching, access-request polling) in follow-up PRs

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -69,15 +69,32 @@ Command Line Options
 
 ### Navigation Data
 
-| Option                            | Description                                      |
-| --------------------------------- | ------------------------------------------------ |
-| `-n, --navigation-address <ADDR>` | Navigation service address for GPS/heading data  |
-|                                   | No value: auto-discover via mDNS                 |
-|                                   | Interface name: search mDNS on that interface    |
-|                                   | `udp:ip:port`: listen for UDP broadcasts         |
-|                                   | `tcp:ip:port`: connect to TCP server             |
-| `--nmea0183`                      | Use NMEA 0183 instead of Signal K for navigation |
-| `--pass-ais`                      | Forward AIS targets from Signal K to GUI clients |
+| Option                            | Description                                                                 |
+| --------------------------------- | --------------------------------------------------------------------------- |
+| `-n, --navigation-address <ADDR>` | Navigation service address for GPS/heading data                             |
+|                                   | No value: auto-discover via mDNS                                            |
+|                                   | Interface name: search mDNS on that interface                               |
+|                                   | `tcp:ip:port`: anonymous Signal K TCP stream                                |
+|                                   | `udp:ip:port`: listen for NMEA 0183 UDP broadcasts                          |
+|                                   | `ws:ip:port`: Signal K WebSocket (via discovery)                            |
+|                                   | `wss:ip:port`: Signal K secure WebSocket (requires `--accept-invalid-certs`)|
+| `--nmea0183`                      | Use NMEA 0183 instead of Signal K for navigation                            |
+| `--pass-ais`                      | Forward AIS targets from Signal K to GUI clients                            |
+| `--accept-invalid-certs`          | Accept self-signed TLS certificates when connecting to Signal K via         |
+|                                   | HTTPS/WSS. Required for boat-LAN setups that use self-signed certs.         |
+
+**Note on authentication:** Authenticated Signal K servers can only be reached
+via `ws:` or `wss:`. The plain `tcp:` transport is anonymous-only. Authentication
+support is tracked as follow-up work.
+
+**Note on TLS SNI:** `wss:ip:port` and mDNS-discovered HTTPS Signal K servers
+use the server's IP address (not hostname) during the TLS handshake. Some
+strict reverse proxies (e.g. nginx with `server_name` enforcement and
+TLS SNI matching) reject handshakes that present an IP as SNI. Boat-LAN
+Signal K deployments typically do not enforce this and the handshake proceeds
+because `--accept-invalid-certs` bypasses hostname verification entirely.
+If you run Mayara against a strict cloud-hosted Signal K instance, configure
+the server to accept IP-based SNI or point Mayara at an on-LAN instance.
 
 ### Stationary Installation
 
--- a/src/lib/mod.rs
+++ b/src/lib/mod.rs
@@ -28,6 +28,7 @@ pub mod network;
 pub mod protos;
 pub mod radar;
 pub mod recording;
+pub mod signalk;
 pub mod stream;
 pub mod util;
 
@@ -77,10 +78,19 @@ pub struct Cli {
     #[arg(short, long, default_value_t, value_enum)]
     pub targets: TargetMode,
 
-    /// Set navigation service address, either
-    /// - Nothing: all interfaces will search via MDNS
-    /// - An interface name: only that interface will seach for via MDNS
-    /// - `udp-listen:ipv4-address:port` = listen on (broadcast) address at given port
+    /// Set navigation service address. Accepts either an interface name
+    /// (restricts mDNS to that interface) or `<scheme>:<address>:<port>`.
+    ///
+    /// Schemes:
+    /// - `tcp:` — plain TCP Signal K stream (anonymous only; auth not supported)
+    /// - `udp:` — UDP listener for NMEA 0183 broadcasts
+    /// - `ws:`  — WebSocket via Signal K discovery (anonymous)
+    /// - `wss:` — WebSocket Secure via Signal K discovery; requires
+    ///            `--accept-invalid-certs`
+    ///
+    /// Authenticated Signal K servers can only be reached via `ws:` or `wss:`
+    /// (auth is not yet implemented; tracked as follow-up work). The plain
+    /// `tcp:` transport is strictly for anonymous setups.
     #[arg(short, long)]
     pub navigation_address: Option<String>,
 
@@ -138,6 +148,10 @@ pub struct Cli {
     /// Pass AIS targets from Signal K server to GUI clients
     #[arg(long, default_value_t = false)]
     pub pass_ais: bool,
+
+    /// Accept invalid TLS certificates (e.g. self-signed) when connecting to Signal K
+    #[arg(long, default_value_t = false)]
+    pub accept_invalid_certs: bool,
 
     /// Use emulator radar instead of real radar discovery
     #[arg(long, default_value_t = false)]
--- a/src/lib/navdata.rs
+++ b/src/lib/navdata.rs
@@ -1,5 +1,5 @@
 use atomic_float::AtomicF64;
-use futures_util::future::select_ok;
+use futures_util::{SinkExt, StreamExt, future::select_ok};
 use mdns_sd::{Error, IfKind, ServiceDaemon, ServiceEvent};
 use nmea_parser::*;
 use serde_json::Value;
@@ -19,6 +19,7 @@ use tokio::{io::AsyncBufReadExt, net::UdpSocket, time::sleep};
 use tokio::{io::AsyncWriteExt, net::TcpStream};
 use tokio::{io::BufReader, sync::broadcast::Receiver};
 use tokio_graceful_shutdown::SubsystemHandle;
+use tokio_tungstenite::tungstenite::Message;
 
 use crate::{
     Cli,
@@ -73,6 +74,20 @@ fn set_own_ship_context(context: &str) {
         if guard.is_none() {
             log::info!("Own-ship context set to: {}", context);
             *guard = Some(context.to_string());
+        }
+    }
+}
+
+/// Clear the own-ship context. Called on every Signal K reconnection so a
+/// roam to a different server (with a different vessel URN) cannot silently
+/// misroute AIS traffic to the stale context.
+fn reset_own_ship_context() {
+    if let Some(lock) = OWN_SHIP_CONTEXT.get() {
+        if let Ok(mut guard) = lock.write() {
+            if guard.is_some() {
+                log::debug!("Clearing own-ship context on reconnect");
+                *guard = None;
+            }
         }
     }
 }
@@ -208,9 +223,7 @@ pub(crate) fn set_sog(sog: Option<f64>) {
     }
 }
 
-/// The hostname of the devices we are searching for.
-const SIGNAL_K_SERVICE_NAME: &'static str = "_signalk-tcp._tcp.local.";
-const NMEA0183_SERVICE_NAME: &'static str = "_nmea-0183._tcp.local.";
+const NMEA0183_SERVICE_NAME: &str = "_nmea-0183._tcp.local.";
 
 /// Subscription for own-ship navigation data only
 const SUBSCRIBE_SELF: &'static str = "{\"context\":\"vessels.self\",\"subscribe\":[{\"path\":\"navigation.headingTrue\"},{\"path\":\"navigation.position\"},{\"path\":\"navigation.speedOverGround\"},{\"path\":\"navigation.courseOverGroundTrue\"}]}\r\n";
@@ -219,10 +232,33 @@ const SUBSCRIBE_SELF: &'static str = "{\"context\":\"vessels.self\",\"subscribe\
 const SUBSCRIBE_ALL: &'static str =
     "{\"context\":\"vessels.*\",\"subscribe\":[{\"path\":\"*\"}]}\r\n";
 
+/// A Signal K subscription the transport layer should send in response to an
+/// incoming message. Both TCP and WebSocket receive loops share this decision
+/// logic via `NavigationData::signalk_actions_for_line`.
+#[derive(Clone, Copy, Debug)]
+enum SignalKSubscription {
+    /// Initial own-ship subscription; sent once we receive the hello message.
+    OwnShip,
+    /// All-vessels subscription for AIS forwarding; sent once we know the
+    /// own-ship context (when `--pass-ais` is set).
+    AllVessels,
+}
+
+impl SignalKSubscription {
+    fn payload(self) -> &'static str {
+        match self {
+            Self::OwnShip => SUBSCRIBE_SELF,
+            Self::AllVessels => SUBSCRIBE_ALL,
+        }
+    }
+}
+
 enum ConnectionType {
     Mdns,
     Udp(SocketAddr),
     Tcp(SocketAddr),
+    /// WebSocket connection; bool indicates TLS (wss) vs plain (ws)
+    Ws(SocketAddr, bool),
 }
 
 impl ConnectionType {
@@ -237,10 +273,11 @@ impl ConnectionType {
                     return ConnectionType::Mdns;
                 } else if parts.len() == 2 {
                     if let Ok(addr) = parts[1].parse() {
-                        // Dump if illegal address
                         match parts[0].to_ascii_lowercase().as_str() {
                             "udp" => return ConnectionType::Udp(addr),
                             "tcp" => return ConnectionType::Tcp(addr),
+                            "ws" => return ConnectionType::Ws(addr, false),
+                            "wss" => return ConnectionType::Ws(addr, true),
                             _ => {} // fallthrough to panic below
                         }
                     }
@@ -248,22 +285,23 @@ impl ConnectionType {
             }
         }
         panic!(
-            "Interface must be either interface name (no :) or <connection>:<address>:<port> with <connection> one of `udp` or `tcp`."
+            "Interface must be either interface name (no :) or <connection>:<address>:<port> with <connection> one of `udp`, `tcp`, `ws` or `wss`."
         );
     }
 }
 
-#[derive(Debug)]
+use crate::signalk::WsStream;
+
 enum Stream {
     Tcp(TcpStream),
     Udp(UdpSocket),
+    WebSocket(WsStream, String),
 }
 
 pub(crate) struct NavigationData {
     args: Cli,
     nmea0183_mode: bool,
     pass_ais: bool,
-    service_name: &'static str,
     what: &'static str,
     nmea_parser: Option<NmeaParser>,
 }
@@ -277,7 +315,6 @@ impl NavigationData {
                 args,
                 nmea0183_mode: true,
                 pass_ais,
-                service_name: NMEA0183_SERVICE_NAME,
                 what: "NMEA0183",
                 nmea_parser: Some(NmeaParser::new()),
             },
@@ -285,7 +322,6 @@ impl NavigationData {
                 args,
                 nmea0183_mode: false,
                 pass_ais,
-                service_name: SIGNAL_K_SERVICE_NAME,
                 what: "Signal K",
                 nmea_parser: None,
             },
@@ -341,15 +377,22 @@ impl NavigationData {
         let navigation_address = self.args.navigation_address.clone();
 
         loop {
+            // Clear per-session state so a reconnect to a different Signal K
+            // server cannot inherit a stale own-ship URN from the previous one.
+            reset_own_ship_context();
+
             match self
                 .find_service(&subsys, &mut rx_ip_change, &navigation_address)
                 .await
             {
                 Ok(Stream::Tcp(stream)) => {
                     log::info!(
-                        "Listening to {} data from {}",
+                        "Listening to {} data via TCP from {}",
                         self.what,
-                        stream.peer_addr().unwrap()
+                        stream
+                            .peer_addr()
+                            .map(|a| a.to_string())
+                            .unwrap_or_else(|_| "<unknown>".to_string())
                     );
                     match self.receive_loop(stream, &subsys).await {
                         Err(RadarError::Shutdown) => {
@@ -362,7 +405,14 @@ impl NavigationData {
                     }
                 }
                 Ok(Stream::Udp(socket)) => {
-                    log::info!("Listening to {} data via UDP", self.what);
+                    log::info!(
+                        "Listening to {} data via UDP from {}",
+                        self.what,
+                        socket
+                            .local_addr()
+                            .map(|a| a.to_string())
+                            .unwrap_or_else(|_| "<unknown>".to_string())
+                    );
                     match self.receive_udp_loop(socket, &subsys).await {
                         Err(RadarError::Shutdown) => {
                             log::debug!("{} receive_loop shutdown", self.what);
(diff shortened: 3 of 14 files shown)


dirkwa:
Hi Marc, Kees and I really appreciate your contribution. Well done so far. I just updated the signalk plugin to support the new radar api. 

If i understand your attempt then you want to support here the exact opposite. 

Knowing that most users have auth enabled I am currently unsure which benefit this will bring for mayara atm. I was just fighting loop events on n2k where SK was sending and we saw rebounds from different sources.

Enlighten us, maybe I am just too SK focused and overlooked a valuable use case, thanks.

msallin:
Indeed thats probably gonna be the most prominent use case.

However, MaYaRa currently runs as standalone and there it is necessary to have this data for certain features. Additionally, certain featuers like ARPA target tracking need the heading. Means, the connection to Signalk et al. may only be dropped or neglected if the standalone option isnt for production and when the features are pushed down into the plugin.

Sometimes hard to follow all conversations in all places. So probably I got that completely wrong.. 

Btw. yes, I forgot the auth..

dirkwa:
Thanks for picking this up, I now understand why this is beneficial  — I support the direction and think it's worth landing. A few thoughts on how I'd shape it before it comes out of draft.

### Why WSS matters: it's the only auth-capable Signal K transport

Looking at signalk-server, the TCP delta interface on port 8375 is anonymous-only by design — see [`[src/interfaces/tcp.ts:43](https://github.com/SignalK/signalk-server/blob/master/src/interfaces/tcp.ts#L43)`](https://github.com/SignalK/signalk-server/blob/master/src/interfaces/tcp.ts#L43): the interface refuses to start unless `allowReadOnly == true`. There is no JWT field, no header, no query param. On any normally-secured Signal K server the port simply isn't listening.

That's the real motivation for adding a WebSocket transport: it's the only Signal K transport that can carry authentication. Most boats run Signal K with security enabled, so without WS mayara can't talk to them at all. This PR is necessary, not nice-to-have.

### Frame WS as the recommended path, keep TCP as legacy

I'd land this as **additive** rather than ripping out the existing TCP path. Concretely:

- Keep `_signalk-tcp._tcp` mDNS browsing and the existing TCP receive loop. They work fine for unsecured local servers and people have them in their configs already.
- Add `_signalk-https._tcp` / `_signalk-http._tcp` mDNS browsing and the new WS receive loop alongside.
- **When both resolve at the same time, prefer the WS endpoint.** Gives us "WS-first, TCP-fallback" automatically without removing anything.
- In the README and the `--navigation-address` help text, document `ws://` / `wss://` as the recommended form, mark `tcp:` for Signal K as legacy/anonymous-only, and state clearly: **authenticated Signal K servers require `ws://` or `wss://`**.
- When mayara hits an SK server that requires auth, log a friendly hint: `"Signal K server requires authentication; switch --navigation-address to ws://… or wss://… to enable token request"`.

Net effect: zero breakage for existing users, every new user gets pointed at WS, and the migration happens organically. We can deprecate `tcp:` for SK in a later release once we know nobody depends on it.

`-n` / `--navigation-address` after this PR:

|`-n` value|Behavior|
|---|---|
|_(nothing)_|mDNS on all interfaces — kept (now browses both `_signalk-tcp` and `_signalk-https`/`_signalk-http`, prefers WS)|
|`<interface>`|mDNS on that interface — kept, same WS-preference rule|
|`tcp:host:port`|Plain TCP — kept; legacy, anonymous-only|
|`udp:host:port`|UDP listen for SK delta broadcasts — kept|
|`ws:host:port`|Plain WebSocket — added in this PR|
|`wss:host:port`|TLS WebSocket — added in this PR|

NMEA 0183 mode (`--nmea0183`) is unaffected — different protocol, different mDNS service, different code path.

### Slim down the dep weight

A few things I'd push back on in the current diff — they're addressable without losing functionality:

1. **Drop `reqwest`.** The HTTP calls (`GET /signalk` discovery, and the access-request POST + poll once we add auth) are three trivial JSON requests to the same host we're about to open a WebSocket against. Hand-writing them on `tokio` + `tokio-rustls` (already in mayara's tree at [[Cargo.toml:83](https://github.com/MarineYachtRadar/mayara-server/blob/main/Cargo.toml#L83)](https://github.com/MarineYachtRadar/mayara-server/blob/main/Cargo.toml#L83)) is ~80 lines total and zero new top-level deps.
2. **Drop `rustls-native-certs`.** Boat-LAN deployments use self-signed certs almost universally. For v1 I'd support exactly two modes: plain `ws://` and `wss://` with `--accept-invalid-certs`. Users with real CA-signed SK certs can still pass `--accept-invalid-certs` as a no-op. We can revisit native roots only if a real user actually needs them.
3. **Inline the `NoVerifier`.** The standalone struct is ~50 lines for what `rustls::ClientConfig::builder().dangerous().with_custom_certificate_verifier(...)` does inline in ~10 lines, reusing `tokio-rustls`.
4. **Keep the `GET /signalk` discovery call** — it's only ~30 lines once `reqwest` is gone, it's spec-correct, and it handles non-default WS paths. Worth keeping.

Net dep diff after this should be: nothing new beyond what `tokio-tungstenite`'s rustls feature pulls in (and later `uuid` for the auth piece).

### Auth: what I'd add on top

I think it's worth landing the access-request flow in this same PR (or a stacked one immediately after), since the WS work only really pays off with auth. Sketch:

- New `src/lib/signalk_auth.rs` (~150 lines) with a `SignalKAuth` struct holding a persistent `clientId` (v4 UUID, generated once) and a `HashMap<server_origin, jwt>` token cache, persisted to `~/.config/mayara/signalk-tokens.json` via the existing `directories` crate.
- Auth lives entirely on the WS path. The TCP path stays auth-free, as it was.
- On WS upgrade 401: spawn a background task that does `POST /signalk/v1/access/requests` with `permissions: "readonly"`, polls `GET <href>` every 10s, and on approval saves the token and reconnects. Non-blocking — mayara starts normally without nav data and recovers when the user approves.
- One `INFO` log line at the start of the request pointing at `https://<host>/admin/#/security/access-requests` so the user knows where to click.
- Reuse the same `clientId` across requests so revoke/re-approve cycles show as one identity in the SK admin UI, not many.
- Two CLI flags: `--signalk-token <jwt>` (manual override) and `--signalk-client-id <uuid>` (pin identity, for moving an approved install).
- Adds one small dep: `uuid = { version = "1", features = ["v4", "serde"] }`.

Happy to pair on any of this, or to take the auth piece as a stacked PR after this one lands cleaned up — your call on how to split the work. Either way I think this is the right direction and worth finishing.

dirkwa:
See also https://github.com/MarineYachtRadar/mayara-server/issues/69
we will need a READ/WRITE token for this.

msallin:
@dirkwa 

**Discovery**
the only problem (maybe only a theoretical one tbh) I see is keep _signalk-tcp._tcp. The reason is that this endpoint is versionized. So when it changes, the integration will break. When doing the discovery flow, we handle that gracefully.

SignalK could expose the discovery endpoint also for unauthenticated clients. That way we could call that endpoint (even if auth is on) and get the tcp endpoint from there (given its activated). 

The only security risk is that the discovery endpoint shows the signalk version. But.. Well.. I dont see that as problematic.

**Auth**

I suggest to first bring that PR in, then add auth afterwards. I suggest going the same path as I did in https://github.com/msallin/signalk-ha. This is the easiest way for non-tech users. Btw brings me back to https://github.com/SignalK/signalk-server/issues/2244


WDYT?

dirkwa:
I will look at it soon. Can you leave draft mode? Thanks 🙏

dirkwa:
Thanks for picking this up, Marc. I spent some time reading through the diff against main and I think the direction is right — this needs to land. A few things I'd like to see changed before it comes out of draft, and a couple of smaller notes.

### Blocking

**1. Please don't delete the `_signalk-tcp._tcp` mDNS path.**

In the current diff [`[src/lib/navdata.rs:212](https://github.com/MarineYachtRadar/mayara-server/blob/main/src/lib/navdata.rs#L212)`](https://github.com/MarineYachtRadar/mayara-server/blob/main/src/lib/navdata.rs#L212) loses the `SIGNAL_K_SERVICE_NAME = "_signalk-tcp._tcp.local."` constant and [`[src/lib/signalk.rs:22-23](https://github.com/MarineYachtRadar/mayara-server/pull/43/files)`](https://github.com/MarineYachtRadar/mayara-server/pull/43/files) only browses `_signalk-http._tcp` and `_signalk-https._tcp`. That's a silent regression for every unsecured Signal K install on upgrade — their mDNS record simply stops being seen.

I'd make it additive: browse `_signalk-tcp._tcp` **and** `_signalk-http(s)._tcp` concurrently, and when a server resolves on both, prefer WS. That gives us WS-first-with-TCP-fallback without removing anything, and leaves the radar-provider-plugin path (local LAN, often anonymous) intact. The version-negotiation argument for the discovery protocol is valid — but it's a reason to _use_ HTTP discovery when we find an HTTP(S) record, not a reason to stop browsing the TCP record.

**2. Flip the transport preference to WS-first.**

[`[src/lib/signalk.rs:211](https://github.com/MarineYachtRadar/mayara-server/pull/43/files)`](https://github.com/MarineYachtRadar/mayara-server/pull/43/files) tries TCP first when discovery returns both endpoints. That's backward for this PR's own motivation: the whole reason to add WS is that it's the only Signal K transport that can carry authentication. Once auth lands on top of this, TCP-first means authenticated servers will never exercise the WS code path in the common case. Please invert: WS first, TCP fallback.

**3. Drop `reqwest` as a runtime dep.**

[`[Cargo.toml:68](https://github.com/MarineYachtRadar/mayara-server/pull/43/files)`](https://github.com/MarineYachtRadar/mayara-server/pull/43/files) adds `reqwest = "0.12"` for what is ultimately three trivial JSON calls (`GET /signalk` now, and access-request `POST` + poll once auth lands). Mayara's runtime tree already has `rustls 0.23` ([`[Cargo.toml:52](https://github.com/MarineYachtRadar/mayara-server/blob/main/Cargo.toml#L52)`](https://github.com/MarineYachtRadar/mayara-server/blob/main/Cargo.toml#L52)) and `tokio-rustls 0.26` ([`[Cargo.toml:80](https://github.com/MarineYachtRadar/mayara-server/blob/main/Cargo.toml#L80)`](https://github.com/MarineYachtRadar/mayara-server/blob/main/Cargo.toml#L80)), and this PR enables `rustls-tls-native-roots` on `tokio-tungstenite` anyway. Adding `reqwest` pulls `hyper-rustls`, `h2`, etc. and duplicates the TLS machinery.

Hand-writing the HTTP over `tokio` + `tokio-rustls` is ~80 lines. The `rustls::ClientConfig` you're already building at [`[src/lib/signalk.rs:287-315](https://github.com/MarineYachtRadar/mayara-server/pull/43/files)`](https://github.com/MarineYachtRadar/mayara-server/pull/43/files) can be reused directly. There's prior art for the builder pattern at [`[src/bin/mayara-server/web.rs:88-100](https://github.com/MarineYachtRadar/mayara-server/blob/main/src/bin/mayara-server/web.rs#L88-L100)`](https://github.com/MarineYachtRadar/mayara-server/blob/main/src/bin/mayara-server/web.rs#L88-L100) (server side but same shape). `reqwest` stays in `[dev-dependencies]` and `[build-dependencies]` — those don't hit the runtime binary.

**4. Drop `rustls-native-certs`.**

[`[Cargo.toml:70](https://github.com/MarineYachtRadar/mayara-server/pull/43/files)`](https://github.com/MarineYachtRadar/mayara-server/pull/43/files) adds it for the wss:// path at [`[src/lib/signalk.rs:287-315](https://github.com/MarineYachtRadar/mayara-server/pull/43/files)`](https://github.com/MarineYachtRadar/mayara-server/pull/43/files). Boat-LAN SK deployments use self-signed certs almost universally. For v1 I'd support exactly two modes: plain `ws://`, and `wss://` with `--accept-invalid-certs`. Users with real CA-signed SK certs still pass `--accept-invalid-certs` as a no-op. We can revisit native roots only if a real user actually needs them.

This also lets `NoVerifier` shrink from the standalone 50-line struct at [`[src/lib/signalk.rs:318-347](https://github.com/MarineYachtRadar/mayara-server/pull/43/files)`](https://github.com/MarineYachtRadar/mayara-server/pull/43/files) to ~10 lines inlined via `rustls::ClientConfig::builder().dangerous().with_custom_certificate_verifier(...)`.

### Non-blocking

**5. Keep the `GET /signalk` discovery.** Once `reqwest` is gone it's only ~30 lines, it's spec-correct, and it handles non-default WS paths. The 14 unit tests in [`[src/lib/signalk.rs:349-607](https://github.com/MarineYachtRadar/mayara-server/pull/43/files)`](https://github.com/MarineYachtRadar/mayara-server/pull/43/files) are good — keep them. Please also add one test asserting both `_signalk-tcp._tcp` and `_signalk-http(s)._tcp` are browsed, and one asserting WS-first preference when discovery returns both.

**6. Auth in a stacked PR.** Agreed with your plan to land this PR first and add access-request auth on top. For that next PR, I'd sketch it as: a `signalk_auth.rs` module holding a persistent `clientId` (v4 UUID, generated once) and a `HashMap<origin, jwt>` token cache, persisted to `~/.config/mayara/signalk-tokens.json` via the `directories 5.0.1` crate we already have ([`[Cargo.toml:43](https://github.com/MarineYachtRadar/mayara-server/blob/main/Cargo.toml#L43)`](https://github.com/MarineYachtRadar/mayara-server/blob/main/Cargo.toml#L43)). Non-blocking 401-retry with access-request POST + 10s poll, one INFO log line pointing at `https://<host>/admin/#/security/access-requests` so the user knows where to approve. Happy to take that as a stacked PR once this one is merged.

**7. When this leaves draft: the test plan checkboxes.** AGENTS.md requires all checkboxes in a PR body to be ticked before review. The five manual cases in the current description need to actually be run (or the list removed).

**8. Friendly 401 log line.** When a WS upgrade comes back 401, log once at INFO: `"Signal K server requires authentication; switch --navigation-address to ws://<host>:<port> or wss://... to enable token request"`. Cheap now, makes the auth PR discoverable.

**9. CLI help text and docs.** [`[src/lib/mod.rs:71-76](https://github.com/MarineYachtRadar/mayara-server/blob/main/src/lib/mod.rs#L71-L76)`](https://github.com/MarineYachtRadar/mayara-server/blob/main/src/lib/mod.rs#L71-L76) should document the full `tcp:`/`udp:`/`ws:`/`wss:` matrix and call out that **authenticated Signal K servers require `ws://` or `wss://`** and `tcp:` is anonymous-only. README/USAGE.md probably need the same update.

**10. Commit hygiene.** Per AGENTS.md the commit body shouldn't carry `Co-Authored-By: Claude` and the PR description shouldn't have the `🤖 Generated with Claude Code` footer. Please strip both before review. PR title is fine.

Thanks again — happy to pair on any of the blocking pieces (hand-rolled HTTP, additive mDNS browsing) if that'd unblock you. Otherwise I'll wait for a push.

msallin:
Thanks for the thorough review, @dirkwa — exactly the kind of feedback I was hoping for on a draft. I've pushed a full rework; the 4 blockers are addressed as follows.

**1. `_signalk-tcp._tcp` preserved.** Now browsing all three services concurrently in `signalk::find_mdns_service`. Anonymous TCP-only installs continue to work without any regression. The new `Transport::Tcp` enum variant takes the direct connect path (no discovery).

**2. WS-first preference.** `connect_via_discovery` now tries `signalk-ws` from the discovery response first and only falls back to `signalk-tcp` on WebSocket failure. Authenticated servers will land on the WS code path in the common case.

**3. `reqwest` removed from runtime deps.** The `GET /signalk` request is now hand-rolled on `tokio` + `tokio-rustls` in `http_get_signalk` (~60 lines). Reuses the same `rustls::ClientConfig` the WebSocket path builds. `reqwest` remains in `[dev-dependencies]` / `[build-dependencies]` only.

**4. `rustls-native-certs` dropped.** `tokio-tungstenite` now uses the `rustls-tls-webpki-roots` feature (no system CA reading), and `NoVerifier` is used for both HTTPS discovery and WSS connections. V1 supports exactly the two modes you described: plain `ws://` and `wss://` with `--accept-invalid-certs`. Trying `wss://` without the flag returns a clear error.

Non-blocking items also done:

- 6 new tests added (additional HTTP parsing, mDNS regression guard for the TCP service constant, and a dedicated test asserting `_signalk-tcp._tcp` remains in the constants list). 27 unit tests total in the new module.
- Friendly 401 log line with a link back to #42 in `connect_websocket`.
- CLI help in `src/lib/mod.rs` and `USAGE.md` rewritten to document the full `tcp:/udp:/ws:/wss:` matrix and call out that authenticated servers require `ws:` / `wss:`.
- Commit hygiene: amended + force-pushed, `Co-Authored-By` trailer removed, PR description footer removed.
- Test plan checkboxes still unchecked — I'll tick them once I've run the manual scenarios locally against my OpenPlotter setup and a TCP-only fallback.

Auth stack agreed — I'll open a follow-up PR with the `signalk_auth.rs` module, persistent `clientId`, and 401-retry flow once this lands. The 401 log line is already there as a breadcrumb.

Leaving as draft until you've had a chance to confirm the approach is what you want. Happy to pair on anything; otherwise I'll wait for a push.

dirkwa:
Forgot to mention, CHANGELOG is now auto-generated. Just don't touch it.

dirkwa:
And please, if possible leave draft mode.

msallin:
@dirkwa okay. Will travel to the Bahamas and be back in a few days to work on this. Feel free to continue if you wanna do. Else, I will take it up again.

msallin:
Pushed another round of review feedback (self-imposed via a panel of specialist reviewers covering Rust idioms, Signal K protocol compliance, performance, maintainability, and functional correctness). Summary below.

### Must-fix (all addressed in the same commit)

- **`RadarError::SignalK(String)` variant** — replaces prior `RadarError::ParseJson` misuse for TLS handshakes, HTTP parse errors, WebSocket failures, config validation, and discovery timeouts. `ParseJson` now means _actually JSON_. (\`src/lib/radar/mod.rs\`, \`src/lib/signalk.rs\`, \`src/lib/navdata.rs\`)
- **Cache `rustls::ClientConfig`** — built once via \`OnceLock<Arc<ClientConfig>>\` in \`signalk::insecure_tls_config\` and shared by both the HTTPS discovery path and the WSS WebSocket path. Was previously rebuilt on every connection.
- **Bound the discovery HTTP read** — \`read_bounded\` wraps the stream in \`AsyncReadExt::take(64 KiB)\` before \`read_to_end\`, so a misbehaving server cannot OOM the Raspberry Pi.
- **Extract shared subscription decision logic** — new \`SignalKSubscription\` enum (\`OwnShip\` / \`AllVessels\`) and \`NavigationData::signalk_actions_for_line\` method de-duplicate hello-detection and AIS-expansion across the TCP and WebSocket receive loops. Replaces \`send_subscription\` + \`send_ais_subscription\` with one \`send_subscription_tcp\` that takes the enum.

### Follow-ups also landed here (same files the PR already touches)

- **Multi-value delta parsing** — \`parse_signalk\` now iterates every \`updates[]\` and every \`values[]\` entry rather than only \`updates[0].values[0]\`. Previously a delta bundling heading + position + SOG + COG in one message (common) would silently drop all but the first field. Also extracted \`apply_signalk_value\` for readability.
- **Reset own-ship context on reconnect** — new \`reset_own_ship_context\` called at the top of \`run()\`'s outer loop so roaming to a different Signal K server cannot inherit a stale vessel URN and misroute AIS updates.
- **Replace \`.peer_addr().unwrap()\` / \`.local_addr().unwrap()\`** in logging paths with \`.map(|a| a.to_string()).unwrap_or_else(...)\`; socket address lookup failures now log \`<unknown>\` instead of panicking.
- **Document TLS SNI limitation** — added a note in USAGE.md explaining that IP-based SNI may be rejected by strict cloud-hosted Signal K servers; boat-LAN setups are unaffected because \`--accept-invalid-certs\` bypasses hostname verification.

### Punted to separate issues

- #139 — typed Serde deserialization for Signal K deltas (hot-path perf; needs a careful look at AIS forwarding's use of the raw \`updates\` \`Value\`).
- #140 — pin \`tokio-tungstenite\` / \`tungstenite\` git forks to a specific \`rev\` for reproducible builds (pre-existing; not introduced by this PR).

### Dismissed after cross-review

- Claimed off-by-one in \`parse_http_body\` at \`header_end + 4\`: not a bug — \`find("\r\n\r\n")\` guarantees that offset is in-bounds; empty body is handled correctly.
- Moving the 27 unit tests to a separate file: inconsistent with the project's inline \`#[cfg(test)] mod tests\` convention (14 other files follow it).
- Fallback TLS without SNI: using \`NoVerifier\`, so SNI mismatches don't affect verification; note added to USAGE.md instead.

Still draft. Happy to pair if any of the above choices don't match your intent.

msallin:
@dirkwa ready to review

dirkwa:
Thanks, can you leave draft mode here? Thanks 🙏

PR #45, opened 2026-04-05
fix(recordings): add missing upload endpoint

msallin:
## Summary

The GUI POSTs uploaded `.mrr`/`.mrr.gz` files to `/recordings/files/upload`, but no handler for that route existed. The request hit axum's default 2 MiB body limit and the connection was torn down mid-upload (HAR showed status 0 after ~540 KB of a 5 MB file).

Add a streaming upload handler that reads the filename from `Content-Disposition`, writes to a temp file, and atomically renames on success. Body limit is raised to 2 GiB on this route only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored GUI recording file upload.
  * Enforces a 2 GiB upload limit and aborts oversized transfers.
  * Validates uploaded filenames and restricts allowed file extensions; rejects invalid uploads with clear errors.
  * Returns a conflict response for duplicate recordings.
  * Improves error handling for failed uploads and returns a success response with filename and size on completion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ Sections can be: Added Changed Deprecated Removed Fixed Security.
 - GUI showed DISCONNECTED in Firefox after server restart due to unnecessary state reset when spoke stream reconnect triggered a redundant state stream reconnect
 - Heading extracted from spoke data for GUI when no external heading source available
 - Furuno spoke data sockets retry on failure instead of silently staying dead
+- Recording file upload from the GUI (add missing `POST /recordings/files/upload` route)
 - Accept 0xc2 as valid Navico spoke status for HALO20+ compatibility (#27)
 - Move inline `display: none` style to CSS for WebGPU warning element
 - `NoSuchRadar` returns 404 (was 500), response includes list of valid radar IDs
--- a/src/bin/mayara-server/web/recordings.rs
+++ b/src/bin/mayara-server/web/recordings.rs
@@ -3,11 +3,13 @@
 use axum::{
     Json,
     body::Body,
-    extract::{Path, Query, State},
-    http::StatusCode,
+    extract::{DefaultBodyLimit, Path, Query, State},
+    http::{HeaderMap, StatusCode},
     response::IntoResponse,
     routing::{delete, get, post, put},
 };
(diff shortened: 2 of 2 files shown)


dirkwa:
Hey Marc, nice PR — this fills a real gap and the streaming approach is solid. A few things I spotted:

**Bug: `.mrr.gz` temp path is wrong**

`target_path.with_extension("mrr.upload")` only replaces the _last_ extension. So `foo.mrr.gz` becomes `foo.mrr.mrr.upload` instead of `foo.mrr.gz.upload`. Two uploads with similar names could also collide on the temp path. Fix is to append `.upload` to the full filename:

```rust
let tmp_name = format!("{}.upload", filename);
let tmp_path = target_path.with_file_name(&tmp_name);
```

**Missing fsync before rename**

`flush()` pushes data to the OS buffer cache but doesn't guarantee it hits disk. If the machine loses power between flush and rename, you could end up with a renamed but empty/corrupt file. Add `file.sync_data().await` after the flush — especially important for multi-GB recordings.

**Orphaned `.upload` temp files**

If the server crashes or client disconnects mid-upload, stale `.upload` files stick around and block future uploads of the same name (the `create_new` returns 409). Fine to defer cleanup to a follow-up, but worth opening an issue so it doesn't get forgotten.

**Minor: TOCTOU on `target_path.exists()`**

There's a small window between the exists check and the final `rename()` where another process could create the target file, and rename would silently overwrite it. Low risk for manual GUI uploads, but if you want to be thorough you could use `renameat2` with `RENAME_NOREPLACE` or a hard-link-then-remove pattern.

**PR hygiene**

The test plan has unchecked `[ ]` items — either check them off once tested or remove them before merge.

Everything else looks good — error cleanup is thorough, `create_new(true)` on the temp file is a nice touch, and keeping the scope tight was the right call. The directory endpoints and download compression observations are good follow-up candidates.

msallin:
@dirkwa thx for review!

Im used to always rebase/squash. @keesverruijt told me not to do so, since we had some conflicts in a PR. Probably a good middleground would be: Dont do when "Allow edits and access to secrets by maintainers" is allowed wdyt?

Will do that here now.

PR #100, opened 2026-04-12
fix(furuno): handle 0x7D as radar alarm, not DRS4W heartbeat

dirkwa:
## Summary

The `0x7D` notification is not a DRS4W-specific heartbeat — it is the general radar alarm notification used across all Furuno models, with wire grammar `$N7D,<type>,<d1>,<d2>,<d3>` and an idle value of `$N7D,0,0,0,0`. Because idle traffic looks periodic, the misinterpretation has gone unnoticed: real alarms with a nonzero type have been silently dropped by the existing no-op match arm in `report.rs`.

Rename the `CommandId` variant to `Alarm` and add a report handler that logs nonzero alarms at warn level and clear transitions at info level, matching the existing `GuardStatus` pattern.

## Verified

- `cargo build --all` and `cargo test` pass
- Runs cleanly against DRS4D-NXT with no regression (the DRS4D-NXT does not emit `$N7D` in practice, so the warn path is untested on hardware — it requires a FAR-class radar to exercise)

## Overview

This PR corrects handling of the Furuno 0x7D notification. The opcode is the general radar alarm notification (wire grammar `$N7D,<type>,<d1>,<d2>,<d3>`, idle `$N7D,0,0,0,0`), not a DRS4W-specific heartbeat. Real alarms with nonzero type values are now handled instead of being silently dropped.

### Protocol Definition
- src/lib/brand/furuno/protocol.rs
  - Renames `CommandId::DRS4WHeartbeat = 0x7D` to `CommandId::Alarm = 0x7D` and updates documentation to describe the `$N7D` alarm message format and idle semantics.

### Report Handler
- src/lib/brand/furuno/report.rs
  - Adds `alarm_active: bool` to `FurunoReportReceiver`, initialized to `false`.
  - Implements handling for `CommandId::Alarm`:
    - Interprets `numbers[0] != 0` as alarm activation and `numbers[0] == 0` as alarm clear.
    - Logs transitions only (avoids repeated logs for unchanged state).
    - Emits warn-level log on transition to an active alarm with alarm type and details (`numbers[1..=3]`).
    - Emits info-level log on alarm clear transition.
  - Removes the previous no-op handling of the 0x7D opcode so alarms are no longer dropped.

## Verification

- `cargo build --all` and `cargo test` pass.
- Runs cleanly against DRS4D-NXT with no regression.
- Note: the warn path requires a FAR-class radar to exercise because DRS4D-NXT does not emit `$N7D` in practice.

### DIFF
--- a/src/lib/brand/furuno/protocol.rs
+++ b/src/lib/brand/furuno/protocol.rs
@@ -332,8 +332,9 @@ pub enum CommandId {
     TuneIndicator = 0x76,
     /// `0x77` — No-transmit sector (sector blanking).
     BlindSector = 0x77,
-    /// `0x7D` — DRS4W-specific heartbeat (~1 Hz).
-    DRS4WHeartbeat = 0x7D,
+    /// `0x7D` — Radar alarm: `$N7D,<type>,<d1>,<d2>,<d3>`.
+    /// Generic across all Furuno models; idle = `$N7D,0,0,0,0`.
+    Alarm = 0x7D,
 
     /// `0x80` — Attenuation.
     Att = 0x80,
--- a/src/lib/brand/furuno/report.rs
+++ b/src/lib/brand/furuno/report.rs
@@ -75,6 +75,7 @@ pub struct FurunoReportReceiver {
     prev_spoke: [Vec<u8>; 2],
     prev_angle: [u16; 2],
     guard_zone_alarm: [bool; 2],
+    alarm_active: bool,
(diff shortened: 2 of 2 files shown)


dirkwa:
@coderabbitai
Fixed the warn-spam finding in 12bab0a — the warn now only fires on the transition into active alarm state, not on every repeated nonzero $N7D notification.

PR #101, opened 2026-04-12
fix(furuno): handle 0xAF as ARPA alarm, not heartbeat

dirkwa:
## Summary

The `0xAF` notification is not a connection heartbeat — it is the ARPA subsystem alarm/status bitmask `$NAF,<bits>`. No connection-liveness logic depends on it; the real keepalive is `AliveCheck` (`0xE3`) sent periodically from `command.rs::send_report_requests()`. The value `$NAF,256` that made it look periodic is a benign single-bit status indicator.

Rename the `CommandId` variant to `ArpaAlarm` and log the bitmask at debug level so it is observable without polluting the default log stream.

Builds on #100 (includes the `0x7D Alarm` rename from that PR). Once #100 merges, this branch can be rebased to show only the `ArpaAlarm` delta.

## Verified

- `cargo build --all` and `cargo test` pass
- Verified on real DRS4D-NXT hardware: `$NAF,256` → `ARPA alarm status 0x0100` logged at debug level, ~1/sec, no regressions

## Summary

This PR corrects the interpretation of two Furuno radar protocol notifications:

### Changes to protocol opcodes

The `CommandId` enum in `src/lib/brand/furuno/protocol.rs` reclassifies two notification types:

- **`0x7D` (Alarm)**: Renamed from `DRS4WHeartbeat` to `Alarm`, properly identifying it as a generic radar alarm command with format `$N7D,<type>,<d1>,<d2>,<d3>` (idle state: `$N7D,0,0,0,0`). This command applies across all Furuno models, not just DRS4W.

- **`0xAF` (ArpaAlarm)**: Renamed from `Heartbeat` to `ArpaAlarm`, correctly identifying it as the ARPA subsystem alarm/status bitmask with format `$NAF,<bits>`. The periodic observed value `$NAF,256` (0x0100) is a benign single-bit status indicator, not a connection heartbeat.

### Changes to report handling

The `FurunoReportReceiver` in `src/lib/brand/furuno/report.rs` adds:

- A new `alarm_active: bool` field to track radar alarm state transitions.

- **Alarm handler** (`CommandId::Alarm`): Logs at `warn` level when alarm_type is non-zero with full details (type and three parameter fields), sets `alarm_active = true`. Logs at `info` level when alarm clears (type = 0 after being active), sets `alarm_active = false`.

[...]

PR #102, opened 2026-04-12
feat(furuno): add anti-jamming filter control (0xE8)

dirkwa:
## Summary

Expose the Furuno anti-jamming filter (command `0xE8`, `JammingAble`) as a user control. The wire grammar is `$SE8,<value>` and the official Furuno DRS4W app queries `$RE8` at connect time, so this is a standard control on all current Furuno radar models — DRS4W, DRS-NXT, and FAR series — and is registered unconditionally rather than capability-gated.

Exposed as a binary Off/On list control in the Advanced category, following the `NoiseRejection` pattern. An unconditional `$RE8` init query is added to `init()` so the current state is reflected in the UI on connect.

Builds on #100 and #101. Once those merge, this branch can be rebased to show only the anti-jamming delta.

## Verified

- `cargo build --all` and `cargo test` pass
- Verified on real DRS4D-NXT: `$RE8` → `$NE8,0` round-trip confirmed, control appears in the UI Advanced section as "Anti-jamming" with Off/On options
- REST API returns `{"value":0}` at `/controls/antiJamming`

### DIFF
--- a/src/lib/brand/furuno/command.rs
+++ b/src/lib/brand/furuno/command.rs
@@ -208,6 +208,9 @@ impl Command {
         self.send(CommandMode::Request, CommandId::BlindSector, &[])
             .await?; // $R77
 
+        self.send(CommandMode::Request, CommandId::JammingAble, &[])
+            .await?; // $RE8
+
         // STC (Sensitivity Time Control) curves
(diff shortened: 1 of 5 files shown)


dirkwa:
@coderabbitai review

PR #111, opened 2026-04-13
fix(furuno): sqrt echo curve for DRS4W low-power radars

dirkwa:
## Summary

- Apply an 18th-root (gamma 0.056) echo curve for DRS4W and DRS models to fix washed-out display. Analysis of DRS4W pcap data shows 95% of echo values are below 64/252 — with a linear mapping these all land in dim blue. The aggressive curve pushes even weak returns into red, matching the Furuno iOS app's vivid output.
- Implemented as a precomputed 256-entry LUT, rebuilt when model is identified via `$N96` (live) or from the beacon model report name (replay).
- NXT and FAR models keep the linear mapping since their distribution covers the full 0-252 range.
- Add `RadarModel::from_model_name()` to detect model from the 170-byte beacon report string. Replaces the hardcoded `DRS4DNXT` default in pcap replay mode, so all Furuno models get correct settings during replay.

## Tested

- DRS4W pcap replay shows vivid red/green returns instead of dim blue
- Model correctly detected as DRS4W from beacon report during replay
- NXT pcap replay still uses linear mapping (unchanged)
- All existing tests pass

## Description

This PR fixes washed-out display on Furuno DRS4W and DRS low-power radars by applying a strong gamma echo curve instead of a linear mapping. The solution includes:

1. Echo curve mapping: Adds a precomputed 256-entry lookup table (`echo_lut`) that maps raw echo byte values to palette indices. DRS4W and DRS (low-power) models use an aggressive 18th-root gamma curve (gamma ≈ 0.056) to expand weak returns into vivid colors; NXT and FAR models retain a linear mapping.

2. Dynamic model detection: Adds `RadarModel::from_model_name()` to parse the model name from the 170‑byte beacon model report and derive the radar model instead of using a hardcoded replay default. Replay mode detects the model from the beacon name and updates settings accordingly; live mode avoids setting `model_known` prematurely so `parse_modules()` still runs and NXT controls register correctly.

3. Runtime LUT initialization: `FurunoReportReceiver` derives the model from initial control settings, computes `echo_lut` at construction, and rebuilds the LUT when the model is first identified during `parse_modules`. The code logs when the low-power gamma curve is selected and updates log text to reference "gamma" instead of "sqrt".

4. Replay firmware/version constant: Introduces `REPLAY_FIRMWARE_VERSION = "00.00"` and uses it when updating settings for replay-detected models.

## Changes

- src/lib/brand/furuno/protocol.rs
  - Adds `pub(crate) fn RadarModel::from_model_name(&str) -> RadarModel` to map beacon model name strings to radar model types with prioritized substring matching (handles variants like "DRS4D-NXT" vs "DRS4DNXT" and FAR variants).

- src/lib/brand/furuno/report.rs
  - Adds `echo_lut: [u8; 256]` to `FurunoReportReceiver`, computed from `info.pixel_colors()` and a `low_power` flag.
  - Introduces `build_echo_lut(pixel_colors, low_power) -> [u8; 256]` producing either the aggressive gamma curve for low-power models or a linear mapping for others.
  - Derives `model` from `info.controls.model_name()` at construction and sets `model_known` only when replay-mode supplies a non-`Unknown` model; recomputes `echo_lut` when `parse_modules` assigns the model during frame processing and logs selection of the gamma curve.
  - Replaces per-spoke inline echo-to-palette computation with direct LUT indexing in `add_spoke_to_common` and threads `echo_lut` through `process_frame`/`process_tile_frame` for both Range A and Range B.

- src/lib/brand/furuno/mod.rs
  - Removes hardcoded `RadarModel::DRS4DNXT`/firmware for replay and instead derives the replay model via `RadarModel::from_model_name(beacon_model)`, calling `settings::update_when_model_known(..., model, "00.00")`. Updates replay logging to indicate the model is detected from the beacon string. Applies same behavior for secondary Range B replay path.

- Misc
  - Prevents setting `model_known = true` from beacon name in live mode to ensure `parse_modules()` runs.
  - Adds `REPLAY_FIRMWARE_VERSION` constant used for replay-mode settings updates.
  - Tests and replay verification: DRS4W pcap replay shows vivid red/green returns; model detection from beacon report works in replay; NXT replay remains unchanged; existing tests pass.

## Impact

The gamma echo curve remaps the many low raw echo values (≈95% below 64/252 in DRS4W captures) into higher palette indices, dramatically improving visibility compared to linear mapping while preserving linear behavior for models whose echo distributions span the full range.

### DIFF
--- a/src/lib/brand/furuno/mod.rs
+++ b/src/lib/brand/furuno/mod.rs
@@ -121,9 +121,9 @@ impl FurunoLocator {
 
             info.start_forwarding_radar_messages_to_stdout(&subsys);
 
-            if self.args.is_replay() {
-                // Detect model from the original beacon model string, not from
-                // controls which persistence may have overwritten.
+            // In replay mode, detect model from the original beacon string
+            // (not from controls which persistence may have overwritten).
+            let replay_model = if self.args.is_replay() {
                 let model = RadarModel::from_model_name(beacon_model);
                 log::info!(
                     "{}: Radar model {} detected for replay mode (from beacon {:?})",
@@ -135,7 +135,10 @@ impl FurunoLocator {
                     settings::update_when_model_known(&mut info, model, REPLAY_FIRMWARE_VERSION);
                     radars.update(&mut info);
                 }
-            }
+                Some(model)
+            } else {
+                None
+            };
 
             // Register and configure Range B if this is a dual-range model
             let mut info_b = info_b.and_then(|ib| radars.add(ib));
@@ -143,8 +146,7 @@ impl FurunoLocator {
                 ib.send_command_addr.set_port(port);
                 ib.report_addr.set_port(port);
                 ib.start_forwarding_radar_messages_to_stdout(&subsys);
-                if self.args.is_replay() {
-                    let model = RadarModel::from_model_name(beacon_model);
+                if let Some(model) = replay_model {
                     if model != RadarModel::Unknown {
                         settings::update_when_model_known(ib, model, REPLAY_FIRMWARE_VERSION);
                         radars.update(ib);
--- a/src/lib/brand/furuno/mod.rs
+++ b/src/lib/brand/furuno/mod.rs
@@ -17,6 +17,8 @@ mod protocol;
 mod report;
 mod settings;
 
+const REPLAY_FIRMWARE_VERSION: &str = "00.00";
+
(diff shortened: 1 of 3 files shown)


dirkwa:
@coderabbitai full review

dirkwa:
@coderabbitai resume

PR #119, opened 2026-04-15
feat(navdata): seed --static-position for real radars

dirkwa:
## Motivation

`--static-position` was only read by the emulator. A real shore-based radar installation (the intended use of `--stationary`) with no connected Signal K / NMEA navigation source had no way to supply its known fixed position and heading, leaving ARPA target geographic positions and the GUI target overlay (which gates rendering on `navigation.headingTrue`) broken.

## Solution

Seed the navdata atomics (position, heading, SOG, COG) from `--static-position` at the start of `start_session`, then spawn a small task that re-broadcasts the heading every 2s so late-joining GUI clients receive it. Mirrors the emulator's existing `broadcast_heading` pattern.

The emulator already calls `get_static_position()` for its own initial position and is unaffected.

## Tested

- Furuno DRS4D-NXT + `--stationary --static-position=52.367 4.904 0.0`
- GUI now shows position (52°22'03.4" N / 4°54'14.8" E) and `Hdg: 0.0°`
- ARPA targets compute geographic positions correctly
- `cargo test` passes

## Overview
This PR adds support for using the `--static-position` flag with real shore-based (stationary) radar installations that lack a connected navigation data source (Signal K / NMEA). Previously, this flag was only utilized by the emulator.

## Changes
The PR modifies `start_session` to initialize navigation state from `--static-position` when provided:
- Seeds `navdata` atomic values with latitude, longitude, heading (converted to radians), SOG (0.0), and COG (heading in radians)
- Spawns a new periodic subsystem ("Static Navigation") that re-broadcasts heading every 2 seconds until shutdown, ensuring late-joining GUI clients receive the heading information
- This mirrors the existing emulator behavior without affecting its operation, as the emulator already calls `get_static_position()` for its initial position

## Impact
- Enables correct ARPA target geographic computations for stationary radars
- Ensures GUI target overlay displays correctly by providing the required navigation data
- Provides consistent heading broadcast to all clients connecting to the radar session

## Testing
Verified with a Furuno DRS4D-NXT radar using `--stationary --static-position=52.367 4.904 0.0`:
- GUI displays the supplied position (52°22'03.4" N / 4°54'14.8" E)
- Heading displays correctly (0.0°)
- ARPA targets compute geographic positions correctly
- All unit tests pass

PR #120, opened 2026-04-15
fix(arpa): stop flooding broadcast channel with per-blob target updates

dirkwa:
## Scope

**This is an intermediate fix, not a full ARPA overhaul.** It targets the specific crash/flood symptom observed when guard zones are enabled on a Furuno radar with many tracked targets. More ARPA work (target stability, merge behaviour, UI updates) will follow in separate PRs.

Independent of #119: useful on its own — without #119 the GUI may still have heading from another Signal K source.

## Problem

With guard zones enabled, the software ARPA tracker produced 17+ targets per revolution. Every blob match on an existing target sent a separate `SignalKDelta` through the 32-slot broadcast channel. With ~1300 blob candidates per ~1s revolution, the channel overflowed constantly:

```
Error on Control channel: channel lagged by 224
```

The SignalK v2 WebSocket handler treated `Lagged` as a fatal error and disconnected the subscriber (`break Ok(())`). The browser reconnected, got flooded again, disconnected — a 10–20 second crash/reconnect cycle. Targets never stayed on screen long enough to see.

## Change

Three small, related fixes:

1. **Batch target updates per revolution** (`radar/target/manager.rs`, `radar/target/tracker.rs`)
   - `check_revolution` now returns `bool` to signal a revolution boundary
   - Only `Promoted` (target first confirmed as tracking) broadcasts immediately
   - All other active-target state is collected and sent in a single delta once per revolution
2. **Recover from broadcast lag instead of disconnecting** (`web/signalk/v2.rs`)
   - `RecvError::Lagged` is now warn + continue; the spoke stream already handled it this way
3. **Raise channel capacity 32 → 128** (`radar/mod.rs`) as headroom

## Not in this PR

- No tracker/motion model tuning
- No deduplication or lifecycle changes
- No UI changes
- No change to how `Lost`/deletion deltas are sent

## Tested

- Furuno DRS4D-NXT + guard zone enabled, 60+ tracked targets
- WebSocket no longer disconnects; no `Error on Control channel` lines in log
- Target circles + velocity arrows render stably
- `cargo test` passes

### DIFF
--- a/src/bin/mayara-server/web/signalk/v2.rs
+++ b/src/bin/mayara-server/web/signalk/v2.rs
@@ -1258,6 +1258,9 @@ async fn ws_signalk_delta(
                             send_message(socket, sk_delta).await?;
                         }
                     },
+                    Err(broadcast::error::RecvError::Lagged(n)) => {
+                        log::warn!("Control channel lagged by {n} messages, resuming");
+                    }
                     Err(e) => {
                         log::error!("Error on Control channel: {e}");
                         break Ok(());
--- a/src/lib/radar/mod.rs
+++ b/src/lib/radar/mod.rs
@@ -563,6 +563,13 @@ impl Display for RadarInfo {
     }
 }
 
+/// Capacity of the Signal K client broadcast channel. Sized to absorb a
+/// full revolution's worth of target updates plus control value changes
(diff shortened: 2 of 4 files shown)


dirkwa:
@coderabbitai review

Issue #140, opened 2026-04-21
Cargo.toml: pin custom tokio-tungstenite / tungstenite git forks to a specific rev


msallin:
## Problem

`Cargo.toml` depends on two custom git forks without any revision pinning:

```toml
tokio-tungstenite = { git = "https://github.com/keesverruijt/tokio-tungstenite", features = [...] }
tungstenite = { git = "https://github.com/keesverruijt/tungstenite-rs.git", branch = "permessage-deflate", features = [...] }
```

The `tokio-tungstenite` entry has no \`rev\` or \`branch\` and will float to whatever the remote HEAD is at fetch time. This breaks reproducible builds: two CI runs on the same code can produce different binaries, and git history won't explain why.

## Proposal

- Add \`rev = "<full-sha>"\` to the \`tokio-tungstenite\` dep, pinning it to the current working commit.
- Optionally also pin \`tungstenite\` to a \`rev\` instead of \`branch\`.
- If the fork's branch is stable and you expect updates, pair the \`rev\` with a comment explaining the fork's purpose so future maintainers know when it's safe to bump.

## Why separate

Pre-existing in \`Cargo.toml\`, not introduced by #43. Fixing it independently keeps the reasoning separate from Signal K work.

PR #145, opened 2026-04-24
chore(deps): pin tungstenite forks and add permessage-deflate support

keesverruijt:
## Summary

- Switch to Signal's tungstenite fork (pinned at `signal-v0.27.0`) via `[patch.crates-io]` so the entire dependency graph uses a single tungstenite version with extension support
- Replace the old `axum_fix` WebSocket extractor with `axum_extract_ws`, which adds a `.permessage_deflate()` builder method that negotiates compression with the client via `Sec-WebSocket-Extensions` headers
- Compression is only activated when the client offers `permessage-deflate` in the upgrade request

## Test plan

- [x] `cargo check` passes
- [x] Verify WebSocket connections work in Chrome without RSV1 errors
- [x] Verify `Sec-WebSocket-Extensions: permessage-deflate` appears in the 101 response when the client supports it

## Overview

This PR upgrades WebSocket support by pinning Signal's tungstenite fork and introducing per-message deflate (compression) negotiation. The compression feature activates only when clients explicitly offer it via the `Sec-WebSocket-Extensions` header, enabling more efficient WebSocket communication.

### Dependencies
- Updates `tokio-tungstenite` and `tungstenite` to crates.io versions with a `[patch.crates-io]` override that pins both to Signal's fork tags (`signal-v0.27.0` for tungstenite)
- Enables `deflate`, `handshake`, and `rustls-tls-native-roots` features on the patched tungstenite crate to support compression and enhanced TLS configuration

### WebSocket Infrastructure
- Replaces the `axum_fix` WebSocket extractor with a new `axum_extract_ws` module that provides improved extension handling
- Introduces `WebSocketUpgrade::permessage_deflate()` builder method for configuring compression negotiation
- Adds client deflate offer detection via `client_offers_deflate` field tracking whether `Sec-WebSocket-Extensions: permessage-deflate` appears in the upgrade request
- Conditionally injects `Sec-WebSocket-Extensions: permessage-deflate` in the 101 response only when the client offers it
- Removes the previous boolean `accept_compression()` API in favor of the more granular `permessage_deflate()` method
- Adds `WebSocket` FusedStream trait implementation and public `selected_protocol()` method on the upgrade type

### Compression Behavior
- Negotiation occurs during the WebSocket handshake when both client and server support permessage-deflate
- Compression applies only to message payloads, not the WebSocket frame headers
- Supports both binary and text frames with transparent deflate/inflate

### Handler Updates
- Updates `spokes_handler` and `ws_signalk_delta_shim` to use the new `permessage_deflate()` configuration method instead of the removed `accept_compression(true)` approach

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -787,6 +787,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "843fba2746e448b37e26a819579957415c8cef339bf08564fe8b7ddbd959573c"
 dependencies = [
  "crc32fast",
+ "libz-sys",
  "miniz_oxide",
  "zlib-rs",
 ]
@@ -1477,6 +1478,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "libz-sys"
+version = "1.1.28"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc3a226e576f50782b3305c5ccf458698f92798987f551c6a02efe8276721e22"
+dependencies = [
+ "cc",
+ "pkg-config",
+ "vcpkg",
+]
+
+[[package]]
 name = "linux-raw-sys"
 version = "0.4.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


keesverruijt:
This fixes #140

PR #151, opened 2026-04-24
fix(controls): round instead of truncate in value conversions

### LINKED ISSUE #26: UI/UX "Auto"-mode inconsistent, toggle change value & unclear state
Since its all about the same, small things and probably be fixed together, I create one issue. Let me know if I should spilt it.
1. For me as an user it isnt intuitively clear if "Auto" is on or off
2. When I set "Gain" on a value higher than 40, then I click on "Auto" and again on "Auto" the value gets decreased.
Image

4. There are two different ways that Auto is displayed (compare Sea clutter vs Gain). One shows the slider with "A" the other doesnt.Image

keesverruijt:
## Summary
- Use `f64::round()` instead of truncating casts when converting control values between user-facing (0–100) and wire (0–255) formats
- Fixes the gain value decreasing on each auto-mode toggle (e.g. gain 45 → wire 114 → back to 44)
- Applies to Navico and Raymarine `scale_100_to_byte`, and the shared `Control::set` integer-step rounding

Fixes #26

## Test plan
- [x] All 166 unit tests pass
- [x] Round-trip verification: gain values 1–100 now survive the user→wire→user conversion without loss

### DIFF
--- a/src/lib/brand/navico/command.rs
+++ b/src/lib/brand/navico/command.rs
@@ -87,7 +87,7 @@ impl Command {
         } else if r < 0.0 {
             r = 0.0;
         }
-        r as u8
+        r.round() as u8
     }
 
     fn mod_deci_degrees(a: i32) -> i32 {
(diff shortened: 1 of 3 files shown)


keesverruijt:
This should fix the auto on/off value changes reported in #26.

PR #157, opened 2026-05-03
fix(gui): apply min/max/step to numeric controls

msallin:
## Summary

Numeric controls (antenna height, antenna offsets, etc.) rendered an unconstrained `<input type="number">` with default `step=1` and no bounds, so spinner clicks ignored the radar's actual increment (e.g. 0.1 m for Navico antenna height) and accepted out-of-range values.

This wires the capability's `minValue`/`maxValue`/`stepValue` onto the input so the browser snaps to the right grid and clamps to the supported range.

Also fixes a drift bug in `roundToStep`: the `(value + step/2).toFixed(precision)` formula double-rounds and `roundToStep(0, 0.1)` returned `0.1`, shifting clean values upward each display cycle. Replaced with `Math.round(value/step)*step`.

Reported symptom: antenna height could only be increased, never decreased.

## Summary

This PR applies device-specific min/max/step constraints to numeric input controls in the GUI and fixes rounding drift so numeric controls align to capability-defined value grids and display clean rounded values.

## Changes

- roundToStep — Rounds values to the nearest multiple of stepValue relative to a baseValue (default 0) using Math.round((value - base)/step) * step, returns NaN for non-finite inputs, ignores non-positive stepValue, and chooses display precision from the exact step magnitude (handles scientific notation). This prevents upward drift (e.g., roundToStep(0, 0.1) returns 0) and removes EPSILON/double-rounding heuristics.
- NumericValue builder — Accepts an optional control object (id, name, control = {}) and sets HTML input attributes min, max, and step when control.minValue, control.maxValue, and control.stepValue are finite (and stepValue > 0). oninput/onchange handlers remain to update and validate values.
- buildSingleControl / control wiring — Numeric controls are passed the full capability object so inputs reflect the capability-provided minValue, maxValue, and stepValue. Sector/zone angle rounding and setControlValue now align rounding to control.minValue (base grid) instead of an absolute zero grid.
- UI behavior — Browser-enforced step and bounds make spinner clicks follow device increments (e.g., 0.1 m or 0.1°), clamp typed values to supported ranges, and eliminate display artifacts and upward drift from repeated operations.

## Behavior Improvements / Test notes

- Antenna height spinner steps in device-specific increments and respects reported 0..60 m bounds (example: Navico 0.1 m increments).
- Bearing alignment spinner steps in device-specific increments (example: 0.1°) and accepts negative values where supported.
- Display values round cleanly without upward drift.

## Other

- No exported/public API surface changes.

### DIFF
--- a/web/gui/control.js
+++ b/web/gui/control.js
@@ -185,29 +185,23 @@ function convertControlToUserUnits(id, control) {
 }
 
 /**
- * Rounds a number to a limited number of decimals, for user pleasure.
+ * Rounds a number to the nearest multiple of stepValue and trims trailing
+ * float artifacts (e.g. 0.30000000000000004 -> 0.3) for display.
  */
 function roundToStep(value, stepValue) {
   value = Number(value);
   if (!Number.isFinite(value) || !Number.isFinite(stepValue)) return NaN;
+  if (stepValue <= 0) return value;
 
-  if (Math.abs(stepValue - 0.1) < Number.EPSILON) {
-    return Number((value + stepValue / 2).toFixed(1));
-  }
-  if (stepValue < 0.02) {
-    return Number((value + stepValue / 2).toFixed(2));
-  }
-  if (stepValue <= 1) {
-    return Number((value + stepValue / 2).toFixed(1));
-  }
+  const rounded = Math.round(value / stepValue) * stepValue;
 
-  const scale = 1 / stepValue;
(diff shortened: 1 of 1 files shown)


msallin:
@coderabbitai review

msallin:
@coderabbitai full review

dirkwa:
@coderabbitai full review

PR #158, opened 2026-05-03
fix(gui): convert user units to SI in do_change

msallin:
## Summary

When a user changed a numeric input, `do_change` stored the typed value (in user units, e.g. degrees) in the local update object and then called `setControlValue`, which assumes SI input and re-applies the SI->user conversion. For bearing alignment that turned `30 deg` into `30 * 180/π ≈ 1718` in the displayed input until the next server broadcast. The radar received the right command but the GUI flashed a nonsensical value, making the control look broken.

Convert `user_units` back to SI via `toSI` before writing to `update`, so `setControlValue`'s display path produces the correct user-units value.

Reported symptom: bearing alignment "seems not to do anything".

### DIFF
--- a/web/gui/control.js
+++ b/web/gui/control.js
@@ -31,7 +31,7 @@ export {
 };
 
 import van from "./vendor/van-1.5.2.js";
-import { toUser } from "./units.js";
+import { toSI, toUser } from "./units.js";
 import {
   fetchRadars,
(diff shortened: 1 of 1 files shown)


msallin:
@coderabbitai review

msallin:
@coderabbitai full review

PR #159, opened 2026-05-03
fix(gui): show feedback when heading missing in acquire mode

msallin:
## Summary

Click-to-acquire silently dropped clicks when `trueHeading` was null, which is the default in standalone mode (no SignalK heading subscription). The cursor changed to crosshair, but clicks produced no target and no explanation, making the feature look broken.

Update the existing acquire-target status div with a "No heading data, cannot acquire" message so the user understands why nothing happens. The acquire API still requires a true bearing (computed from relative click + ship heading), so the underlying requirement is unchanged.

Reported symptom: "I can't do anything when I click acquire targets; mouse pointer changes but click has no effect."

This PR fixes a UX issue where target acquisition silently fails when heading data is unavailable. In standalone mode without a SignalK heading subscription, clicking to acquire targets would show the crosshair cursor but produce no target and no explanation, making the feature appear broken.

The fix updates the `#handleMouseUp` method to check heading availability after validating that a click occurred (not a drag). When `trueHeading` is null, the code logs a warning and displays "No heading data, cannot acquire" in the acquire-target status div, providing clear feedback to users about why the acquisition failed. When heading is available, the code proceeds with the existing acquisition logic: computing the true bearing from the relative click angle plus ship heading, normalizing it to [0, 2π), and calling `onTargetAcquire` with the bearing in radians and distance in meters.

### DIFF
--- a/web/gui/ppi.js
+++ b/web/gui/ppi.js
@@ -1763,21 +1763,30 @@ class PPI {
       const clickDistance = Math.sqrt(dx * dx + dy * dy);
 
       // Only treat as click if mouse didn't move much (not a drag)
-      // Require heading data for MARPA acquisition
-      if (clickDistance < 5 && this.trueHeading !== null) {
-        const radarCoords = this.#pixelToRadarCoords(coords.x, coords.y);
-        // Keep bearing in radians, add true heading to get bearing true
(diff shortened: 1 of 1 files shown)


msallin:
@coderabbitai review

msallin:
@coderabbitai full review

PR #167, opened 2026-05-08
fix(furuno): defer spoke decode until Target Analyzer state is known

dirkwa:
## Motivation

On a fresh connection the wire-to-legend LUT defaults to the TA-Off mapping, but an NXT radar can be transmitting in TA-On wire format (`[class:2 | intensity:4 | 00:2]`) before the first `\$NEF` reply arrives. Until that reply lands — up to one report-request interval after model identification — spokes are decoded with the wrong LUT, and high-intensity stationary echoes can be misclassified into Doppler legend slots.

## Fix

Gate spoke emission on a per-range `target_analyzer_known: [bool; 2]` flag that flips to true on:

- the first `\$NEF` for that range, or
- model identification when the model has no Doppler control (so `\$NEF` will never come).

Delta-decoding state (`prev_spoke`, `prev_angle`) keeps updating while the gate is closed so the spoke stream resumes cleanly. Replay mode is treated as known from the start since saved control state already drives the LUT.

## Tested

- DRS4D-NXT 6424A live: server logs `Target Analyzer state confirmed (mode Off), spoke output enabled` ~5s after model identification (one report-request interval). Bug not directly reproduced — would require booting the radar in TA-On mode and reconnecting Mayara before the next report cycle. The logic is straightforward gating around an existing rebuild path.
- `cargo test --lib`: 195/195 pass.
- `cr review --plain`: no findings.

## fix(furuno): defer spoke decode until Target Analyzer state is known

This PR addresses a race condition where spokes are decoded with an incorrect wire-to-legend lookup table on fresh connections to Furuno DRS4x NXT radars.

### Problem

On DRS4x NXT radars, the wire-to-legend LUT defaults to Target Analyzer (Doppler) Off mode. However, the radar may transmit spokes in TA-On wire format before the first $NEF (Target Analyzer state) report arrives—up to one report-request interval after model identification. This causes spokes to be decoded incorrectly, with high-intensity stationary echoes misclassified into Doppler legend slots until the state is confirmed.

### Solution

The PR adds a per-range boolean gate `target_analyzer_known` that defers spoke emission until the Target Analyzer state is confirmed. The gate is set to true when either:

- The first $NEF report arrives for that range, or
- Model identification shows the model has no Doppler control (so $NEF will never arrive)

In replay mode, the gate starts true because saved control state already drives the LUT.

### Implementation

- Added `target_analyzer_known: [bool; 2]` field to `FurunoReportReceiver` (one per range: A and B)
- In `new()`, this is initialized to `true` in replay mode, `false` in live mode
- When processing $NEF reports (`CommandId::TargetAnalyzer`), the corresponding range's flag is set to true with a log message confirming the TA state and spoke output resumption
- During model identification in `$N96` parsing, if the model lacks a Doppler control, the corresponding range's flag is immediately set to true, bypassing the need to wait for $NEF
- In both frame decoders (`process_frame` for IMO and `process_tile_frame` for Tile), `add_spoke_to_common` is only called when `target_analyzer_known[range_idx]` is true
- Delta-decoding state (`prev_spoke`, `prev_angle`) continues updating regardless of gate status, ensuring clean stream resumption

### Testing

Live testing on DRS4D-NXT 6424A confirmed Target Analyzer state logging and spoke output resumption approximately 5 seconds after model identification. All 195 unit tests pass.

### DIFF
--- a/src/lib/brand/furuno/report.rs
+++ b/src/lib/brand/furuno/report.rs
@@ -92,6 +92,13 @@ pub(crate) struct FurunoReportReceiver {
     /// DRS4W/DRS use an 18th-root gamma curve to spread the bottom-heavy
     /// echo distribution; other models use a linear mapping.
     wire_to_legend: [[u8; 256]; 2],
+    /// Whether the Target Analyzer (Doppler) state for each range has been
+    /// confirmed by either a `$NEF` report from the radar or by learning
+    /// that the model has no TA capability. Until this is true, the LUT
+    /// in `wire_to_legend` may not match the radar's actual wire format,
+    /// so we drop spokes rather than render them with wrong colors.
+    /// Index 0 = Range A, 1 = Range B.
+    target_analyzer_known: [bool; 2],
 }
 
 impl FurunoReportReceiver {
@@ -128,6 +135,12 @@ impl FurunoReportReceiver {
             blob_tx,
         );
 
+        // In replay mode the saved control state already drives the LUT, so
+        // treat TA state as known from the start. In live mode we wait for
+        // either a `$NEF` reply or model identification to confirm whether
+        // the radar has TA at all.
+        let initial_ta_known = args.is_replay();
+
         FurunoReportReceiver {
             common,
             common_b: None,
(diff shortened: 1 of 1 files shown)


dirkwa:
@coderabbitai review

PR #168, opened 2026-05-08
feat(recording): faithful playback with source brand identity

dirkwa:
## Motivation

Spokes from a recording rendered with the wrong colors and the playback radar surfaced as a generic \`Brand::Playback\` to consumers — both visible in the GUI and forwarded into Signal K. Goal: record any brand, play it back, and have it look identical to live (PPI rendering, brand exposed via API). The Signal K plugin then picks it up automatically since it just forwards what's on the WebSocket.

## Approach

- New \`RecordingIdentity\` struct captures the source radar's brand, legend-shaping flags (\`doppler_levels\`, \`has_rain_class\`), pixel-value layout, and naming. Serialized as the capabilities JSON of the \`.mrr\` file.
- Player constructs the playback \`RadarInfo\` with the **source brand**, applies the recorded \`set_doppler_levels\` / \`set_has_rain_class\` so the legend matches the wire format the recorded byte indices were mapped against, and sets \`replay = true\` so brand-specific live receivers don't try to attach.
- New \`replay\` field on the \`/radars\` API entry lets the GUI detect playback radars without sniffing the radar key. Filter on \`info.replay()\` instead of \`Brand::Playback\` for "available to record from".
- Bumps \`MRR_VERSION\` to 2; v1 files are rejected with a clear error since they rendered with a generic legend that never matched the source radar.

## Out of scope

Full reproduction of brand-specific controls (Range, Gain, etc.) requires a recorder change to capture control deltas plus a player-side replay scheduler — better as a follow-up. UserName and ModelName are reproduced so the radar list still matches what was recorded.

## Tested

- \`cargo test --lib\`: 196/196 pass (one new test: v1 rejection).
- \`cr review --plain\`: 3 findings on the original commit, all addressed before push (replay doc wording clarified; \`dual\` field added to identity for dual-range A/B preservation; the third was a false positive about a missing PR title).
- Live test pending — the user will validate end-to-end on the DRS4D-NXT 6424A by recording a session and replaying it via the same viewer URL.

## Overview

This PR adds faithful playback of recorded radar sessions with source brand identity preservation. It introduces a `RecordingIdentity` struct to capture source radar metadata in `.mrr` files, updates the recording player to reconstruct radars with their original brand and appearance settings, and exposes a new `replay` field in the radar discovery API for reliable playback detection.

### Recording Identity Capture and Restoration
A new `RecordingIdentity` struct captures the source radar's defining characteristics: brand ID, spoke/pixel layout parameters, legend-shaping flags (doppler_levels, has_rain_class), naming (UserName, ModelName), and dual-range settings. The struct serializes to the capabilities JSON within `.mrr` files and provides JSON conversion methods.

### Recording Playback Enhancement
The recording player now parses `RecordingIdentity` from the `.mrr` capabilities JSON and uses it to configure the playback radar instance. This ensures the playback radar inherits the original source radar's brand identity, spoke and pixel layouts, legend configuration, and UI naming. If the capabilities cannot be parsed, the player logs a warning and falls back to a minimal identity synthesized from the recording header, then to `Brand::Playback` if brand resolution fails. The playback radar is marked with `replay = true` to prevent brand-specific live receivers from attaching.

### API Exposure of Playback Status
The `/signalk/v2/api/vessels/self/radars` endpoint response model gains a new boolean `replay` field indicating whether a radar is sourced from playback. The recordable-radars endpoint now filters radars using the `replay()` getter rather than brand equality against `Brand::Playback`, enabling consistent detection of playback radars.

### Radar Metadata API
`RadarInfo` gains public getter methods (`replay()`, `targets()`, `doppler_levels()`, `has_rain_class()`) and a setter (`set_replay()`) to expose and override runtime radar state. `TargetMode` now derives `Deserialize` in addition to `Serialize` for full serde support.

### MRR File Format Version Bump
The MRR file format is updated to version 2. Version 1 recordings are now rejected with a clear error message, as v1 used a generic legend that did not match source radars. The version check is tightened to reject any file whose header version is not exactly `MRR_VERSION`.

### Client-Side Playback Detection
The GUI's `loadRadar` function now prefers the radar's `replay` flag from fetched metadata, with fallback to the existing `isPlaybackRadar()` check for backward compatibility.

## Testing
Unit tests verify JSON round-trip preservation, optional field handling, forward compatibility with unknown fields, and explicit v1 rejection. Cargo test passes with 196/196 tests.

## Out of Scope
Full replay of brand-specific controls (range, gain, etc.) is deferred to a follow-up requiring recorder-side capture of control deltas and a player scheduler.

### DIFF
--- a/src/bin/mayara-server/web/recordings.rs
+++ b/src/bin/mayara-server/web/recordings.rs
@@ -217,7 +217,7 @@ async fn get_recordable_radars(State(state): State<Web>) -> impl IntoResponse {
     let radars = state.radars.get_active();
     let recordable: Vec<RecordableRadar> = radars
         .iter()
-        .filter(|r| r.brand != mayara::Brand::Playback)
+        .filter(|r| !r.replay())
         .map(|r| RecordableRadar {
             id: r.key(),
             name: r.controls.user_name(),
@@ -268,13 +268,20 @@ async fn start_recording_handler(
         }
     };
 
-    let capabilities_json = serde_json::to_vec(&serde_json::json!({
-        "brand": format!("{}", radar.brand),
-        "spokesPerRevolution": radar.spokes_per_revolution,
-        "maxSpokeLen": radar.max_spoke_len,
-        "pixelValues": radar.pixel_values,
-    }))
-    .unwrap_or_default();
+    // Recording a playback radar would copy the recorded stream back into
+    // a new file. Reject it at the API boundary so callers can't bypass
+    // the GUI filter in get_recordable_radars.
+    if radar.replay() {
+        return (
+            StatusCode::BAD_REQUEST,
+            Json(serde_json::json!({"error": "Cannot record a playback radar"})),
+        );
+    }
+
+    // Capture the source radar's full identity so playback can reproduce
+    // the same legend and brand shape rather than rendering against a
+    // generic Brand::Playback default.
+    let capabilities_json = mayara::recording::RecordingIdentity::from_radar_info(&radar).to_json();
 
     let initial_state = build_initial_state(&radar);
 
--- a/src/bin/mayara-server/web/signalk/v2.rs
+++ b/src/bin/mayara-server/web/signalk/v2.rs
@@ -198,6 +198,12 @@ struct RadarApiV3 {
     /// IP address of the radar unit on the network
     #[schema(value_type = String, example = "192.168.1.50")]
     radar_ip_address: Ipv4Addr,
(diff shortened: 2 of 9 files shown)


dirkwa:
@coderabbitai review

dirkwa:
@coderabbitai review

dirkwa:
@coderabbitai review

dirkwa:
@coderabbitai review

dirkwa:
@coderabbitai review

PR #241, opened 2026-05-25
feat(furuno): draw oversampled spoke samples past the outer range ring

dirkwa:
## Why

Match what the Furuno MFD does: fill the canvas corners outside the outer range ring with the oversampled echo data the radar already transmits.

## How

Furuno's wire format carries `sweep_len` samples per spoke where only the first `scale` cover `0..range_meters`; the rest are real returns from past the configured display range that we previously discarded in `stretch_spoke`. We now keep them and widen the reported spoke range to `range_meters * sweep_len / scale`, which the existing renderer machinery (the emulator already runs at 1.5× spoke_range) scales correctly so the overshoot lands in the canvas corners.

Tile path is unchanged — no overshoot exists in that encoding. The configured display range still drives ring placement, VRM/EBL, guard zones, ARPA targets; only the spoke metadata downstream distance math uses is widened.

## Tested

- Verified on real DRS4D-NXT hardware
- `cargo test --release` all 218 tests pass

## Overview

This PR modifies Furuno spoke handling to preserve and render oversampled radar samples that lie beyond the configured display range. The Furuno wire format provides sweep_len samples per spoke where only the first scale samples correspond to 0..range_meters; the remaining samples represent returns past the configured display range. The change retains these extra samples and widens the reported spoke physical range to range_meters * sweep_len / scale so the existing renderer draws the overshoot into the canvas corners outside the outer range ring.

## Changes

- Adds map_with_overshoot alongside the existing stretch_spoke helper to stretch the full sweep_len buffer into the fixed GUI sample size while computing a widened physical spoke_range based on the overshoot factor.
- Updates IMO frame processing to map decoded spokes using map_with_overshoot; map_with_overshoot falls back to prior behavior when scale is missing, zero, or covers the whole spoke.
- Adds an explicit spoke_range parameter to add_spoke_to_common and threads that widened spoke_range into common.add_spoke(...) so downstream distance/polar math places overshoot samples correctly.
- Keeps Tile-frame (NXT) spoke emission gating unchanged but passes metadata.range as the explicit spoke_range for Tile frames.
- Leaves replay-mode ControlId::Range updates deriving from metadata.range.
- Restores stretch_spoke docblock to its helper and adds four unit tests for map_with_overshoot (one widening case and three fallback cases).

## Scope

Tile encoding/path is unchanged (tiles contain no overshoot). The configured display range continues to drive ring placement, VRM/EBL, guard zones, and ARPA targets; only the spoke metadata used for downstream distance math is widened to enable rendering of oversampled returns.

## Testing

Verified on real DRS4D-NXT hardware. cargo build --release succeeds. Unit test suite grows from 218 to 222 tests and passes. cr review --plain against main produces no findings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/MarineYachtRadar/mayara-server/pull/241?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/brand/furuno/report.rs
+++ b/src/lib/brand/furuno/report.rs
@@ -1274,17 +1274,19 @@ impl FurunoReportReceiver {
             // The GUI buffers each angle in a slot of SPOKE_LEN samples
             // and treats that whole slot as covering the spoke's reported
             // physical range, so sample i is drawn at
-            // `i / SPOKE_LEN * metadata.range`.
+            // `i / SPOKE_LEN * spoke_range`.
             //
             // The radar always transmits `sweep_len` total samples per spoke,
             // but only the first `metadata.scale` of them cover the configured
             // display range (0..range_meters). Samples beyond `scale` are
-            // oversampled data outside the display range.
-            //
-            let send_spoke: Vec<u8> = Self::stretch_spoke(
+            // oversampled returns from past the display range — keep them and
+            // widen the reported spoke range so the renderer can draw them in
+            // the corners outside the outer range ring.
+            let (send_spoke, spoke_range) = Self::map_with_overshoot(
                 &generic_spoke,
                 metadata.scale as usize,
-                SPOKE_LEN,
+                sweep_len,
+                metadata.range,
             );
 
             // Defer emission until the radar's Target Analyzer state is
@@ -1298,6 +1300,7 @@ impl FurunoReportReceiver {
                     Self::add_spoke_to_common(
                         self.common_b.as_mut().unwrap(),
                         &metadata,
+                        spoke_range,
                         angle,
                         heading,
                         &send_spoke,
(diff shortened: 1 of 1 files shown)


dirkwa:
@coderabbitai
Both findings addressed in 864a27a:
- restored the `stretch_spoke` docblock to its own function (it had been orphaned above `map_with_overshoot` after the rename); kept the wording focused on what each helper does
- added four unit tests covering `map_with_overshoot`: one for the widening path, three for the fallback paths (scale == 0, scale >= usable, range == 0)

All 222 tests pass (was 218 + 4 new).

dirkwa:
@coderabbitai
Please run a fresh review on commit 864a27a — it addresses your prior findings on this PR (docblock fix on `stretch_spoke` / `map_with_overshoot`, plus four new unit tests for the widening and fallback branches of `map_with_overshoot`).

dirkwa:
@coderabbitai
Addressed in d27e103 — rephrased the `stretch_spoke` docblock to describe what callers pass without naming a non-existent helper.

PR #243, opened 2026-05-25
fix(gui): round VRM/EBL distance readout to navigator-useful precision

dirkwa:
## Why

The VRM/EBL readout was showing raw float distances like `0.07307722466665459 nm` — looked broken, useless for navigation.

## How

The readout was calling `formatRangeValue`, which is built for radar tick labels that snap to nice fractions (1/4 nm, 1/8 nm, …). For arbitrary VRM distances it falls through to `v / NAUTICAL_MILE + " nm"`, exposing JavaScript's default float-to-string.

Add a dedicated `formatVrmDistance` helper that returns a 2-decimal value in the operator's current unit: `.xx nm` imperial, integer m or `.xx km` metric.

## Tested

- Manually verified the readout now reads e.g. `0.07 nm` / `135 m` / `1.24 km` instead of raw floats

### DIFF
--- a/web/gui/ppi.js
+++ b/web/gui/ppi.js
@@ -23,6 +23,21 @@ function is_metric(v) {
   return divides_near(v, 500);
 }
 
+// Format a distance for a VRM/EBL readout. Unlike `formatRangeValue` (which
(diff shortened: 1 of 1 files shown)


dirkwa:
@coderabbitai review

PR #249, opened 2026-05-25
fix(gui): keep PPI sized to the visible viewport on mobile

### LINKED ISSUE #250: GUI: PPI is mis-sized on mobile browsers when URL bar transitions
Mirror of MarineYachtRadar/mayara-server-signalk-plugin#36 in the right repo (the GUI lives here, not in the plugin).

On Chrome Mobile, `height: 100vh` sizes to the "large viewport height" that includes the URL-bar area, so when the URL bar is visible the PPI container extends below the visible viewport and the radar circle appears compressed into the lower half of the screen. The WebGL/WebGPU canvas backing buffer then drifts out of sync with the overlay canvas (which redraws against a fresh `getComputedStyle`), so the radar image floats above the range rings instead of sitting inside them.

Reported by @Colug0 on Pixel 6 / Android 16 / Chrome 147.

Tracked here so PR #249 can auto-close on merge.

dirkwa:
## Why

On Chrome Mobile, `height: 100vh` sizes to the "large viewport height" that includes the URL-bar area, so when the URL bar is visible the PPI container extends below the visible viewport and the radar circle appears compressed into the lower half of the screen. The WebGL/WebGPU canvas backing buffer then drifts out of sync with the overlay canvas (which redraws against a fresh `getComputedStyle`), so the radar image floats above the range rings instead of sitting inside them.

Reported on Pixel 6 / Android 16 / Chrome 147 in MarineYachtRadar/mayara-server-signalk-plugin#36.

## How

- **CSS**: use `100dvh` (dynamic viewport height) which tracks the visible area as the URL bar collapses or expands. `100vh` is kept as the fallback for browsers without dvh support.
- **viewer.js**: redraw the canvas on `visualViewport.resize` and `screen.orientation.change` (with `orientationchange` fallback) too — Chrome Mobile fires those for URL-bar transitions without firing `window.resize`.

## Tested

- `cr review --plain` against main — findings addressed before push (use `screen.orientation` over deprecated `orientationchange` event; guard for environments where `screen.orientation` exists without a usable `addEventListener`)

Closes #250 (which mirrors MarineYachtRadar/mayara-server-signalk-plugin#36 — the plugin issue will be closed manually after merge).

### DIFF
--- a/web/gui/debug.css
+++ b/web/gui/debug.css
@@ -13,6 +13,7 @@
   right: 0;
   width: 450px;
   height: 100vh;
+  height: 100dvh;
   background: #1a1a2e;
   border-left: 1px solid #333;
   display: flex;
--- a/web/gui/layout.css
+++ b/web/gui/layout.css
@@ -2,10 +2,15 @@
    Layout - Container, Panels, PPI, Canvas
    ============================================ */
 
-/* Main container */
+/* Main container.
+ * `100dvh` (dynamic viewport height) tracks the visible area as Chrome
+ * Mobile's URL bar collapses/expands, so the PPI canvas no longer extends
+ * past the visible viewport and the radar circle stays centred.
+ * `100vh` is the fallback for browsers without dvh support. */
 div.myr_container {
   width: 100%;
   height: 100vh;
+  height: 100dvh;
   overflow: hidden;
 }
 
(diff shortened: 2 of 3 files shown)


dirkwa:
@coderabbitai
Addressed in 8399f06 — added the `typeof screen.orientation.addEventListener === "function"` probe before subscribing; the `window.orientationchange` fallback still kicks in when the method isn't available.

PR #254, opened 2026-05-25
fix(gui): exponential backoff for heading and AIS reconnects

dirkwa:
## Why

When mayara-server was unreachable for a while, the heading and AIS WebSockets retried every 5 seconds with no cap. A 30-minute outage produced 300+ near-identical "reconnecting in 5s" log lines that drowned out any real error.

## How

Apply the same 1s-base / 30s-cap exponential backoff the state stream already uses in `control.js`. Each socket has its own attempts counter so a working AIS socket doesn't reset the heading counter or vice versa. Counters reset on successful `onopen` and (for AIS) when the operator toggles the overlay off, so a real reconnect after an intentional pause still happens fast.

## Tested

## web/gui/viewer.js

Adds exponential backoff for WebSocket reconnections on both heading and AIS streams, replacing the fixed 5-second retry interval. Introduces two reconnection constants (`RECONNECT_BASE_MS = 1000`, `RECONNECT_MAX_MS = 30000`) and a `reconnectDelay(attempts)` helper that calculates exponential delays capped at 30 seconds.

Each socket maintains its own reconnect attempt counter (`headingReconnectAttempts`, `aisReconnectAttempts`). On successful connection (`onopen`), the respective counter resets to 0. On connection loss (`onclose`), the code computes the exponential delay, increments the attempt counter, logs the attempt count and calculated delay, and schedules reconnection after that delay instead of the hardcoded 5 seconds.

The AIS attempt counter also resets when the AIS overlay is manually unsubscribed, allowing fast reconnection after intentional pauses. This keeps heading and AIS socket counters independent so one socket's successful reconnection doesn't reset the other's backoff progress.

[Review Change Stack](https://app.coderabbit.ai/change-stack/MarineYachtRadar/mayara-server/pull/254?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/web/gui/viewer.js
+++ b/web/gui/viewer.js
@@ -35,6 +35,20 @@ import { PPI } from "./ppi.js";
 var webSocket;
 var headingSocket;
 var aisSocket;
+
+// Exponential reconnect backoff for the heading + AIS WebSockets. Caps at
+// 30s so the console doesn't log a reconnect attempt every five seconds
+// while the server is down (we used to log >300 attempts/30min outage).
+// The state stream in control.js has its own equivalent counter — keep
+// them independent so a working AIS socket doesn't reset the others.
+const RECONNECT_BASE_MS = 1000;
+const RECONNECT_MAX_MS = 30000;
+let headingReconnectAttempts = 0;
+let aisReconnectAttempts = 0;
+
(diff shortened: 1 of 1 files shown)


dirkwa:
@coderabbitai review

PR #258, opened 2026-05-25
fix(signalk): require own-ship navigation before trusting an upstream

dirkwa:
## Why

PR #256 cooled down servers that closed without sending any deltas at all, which made mayara settle on the first server that kept its WebSocket open. That's not strong enough: a server can keep the WS open indefinitely while publishing AIS or other vessel data but never producing the own-ship `navigation.position` / `navigation.headingTrue` mayara needs to place AIS targets and rotate the overlay. The result looked like "AIS targets not appearing" even though mayara was happily connected.

## How

- Each receive loop now races a 30-second probe against the message read. If no own-ship nav delta arrives within the window, the loop drops the connection and the caller cools the server down with a distinct "did not publish own-ship navigation" log explaining what to check (security policy / GPS source).
- A first nav delta cancels the probe for the life of the connection.
- One-shot WARN when every Signal K server we've discovered this cycle is in cooldown, so the operator sees mayara isn't silently spinning while waiting for cooldowns to expire. Reset on successful connect so a recovered network produces the warning afresh next time.
- NMEA0183-over-TCP connections skip the probe — they don't go through the Signal K delta path.

## Tested

- New unit test for the no-nav cooldown marker.
- `cargo build --release` clean.
- `cargo test --release` 226 tests pass (was 225 + 1 new).
- `cr review --plain` against main — no findings.

## Changes

Signal K connections now require receipt of own-ship navigation data (`navigation.position` or `navigation.headingTrue`) within 30 seconds to be considered successfully connected.

**src/lib/navdata.rs**: Introduces `SIGNALK_OWN_SHIP_NAV_SEEN` flag to track when own-ship navigation deltas have been applied, separate from the existing `SIGNALK_DELIVERED_DATA` flag. Both TCP and WebSocket Signal K receive loops enforce a 30-second probe deadline that closes the connection if own-ship navigation is not received within the window. Receipt of the first own-ship navigation delta disables the probe for the connection's lifetime. NMEA0183-over-TCP connections bypass the probe entirely. Connection health state is updated to distinguish between servers with no deltas, servers with deltas but no own-ship navigation, and servers providing own-ship navigation.

**src/lib/signalk.rs**: Adds `mark_signalk_server_silent_no_nav()` to cool down servers that connect successfully but don't publish own-ship navigation, using the same cooldown mechanism as the existing silent marker. The mDNS discovery loop now tracks all Signal K socket addresses seen in the current discovery cycle and emits a single warning when every discovered server is in cooldown, listing the affected addresses and providing guidance about configuration (`--signalk-token`) and upstream navigation data sources. This warning resets on successful connection.

## Testing

Adds unit test verifying that the no-nav silent marker shares cooldown tracking with the existing silent marker. All 226 tests pass (225 existing + 1 new).

[Review Change Stack](https://app.coderabbit.ai/change-stack/MarineYachtRadar/mayara-server/pull/258?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/navdata.rs
+++ b/src/lib/navdata.rs
@@ -1242,19 +1242,24 @@ fn apply_signalk_value(values_entry: &Value, source: &str) {
     log::trace!("parse_signalk: path = '{}', value = {:?}", path, value);
     match path {
         "navigation.position" => {
-            set_position(
-                value["latitude"].as_f64(),
-                value["longitude"].as_f64(),
-                source,
-            );
-            // First own-ship position turns the probe deadline off for the
-            // life of this connection — even a single delta tells us the
-            // upstream is publishing what the radar overlay needs.
-            mark_signalk_own_ship_nav_seen();
+            let lat = value["latitude"].as_f64();
+            let lon = value["longitude"].as_f64();
+            set_position(lat, lon, source);
+            // Only arm the probe-satisfied flag for a delta that carries a
+            // real fix. Signal K servers do publish `value: null` (or a
+            // partial position) when the upstream loses GPS — counting
+            // those would keep mayara stuck on a server that isn't
+            // actually providing nav.
+            if lat.is_some() && lon.is_some() {
+                mark_signalk_own_ship_nav_seen();
+            }
         }
         "navigation.headingTrue" => {
-            set_heading_true(value.as_f64(), source);
-            mark_signalk_own_ship_nav_seen();
+            let heading = value.as_f64();
+            set_heading_true(heading, source);
+            if heading.is_some() {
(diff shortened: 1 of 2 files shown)


dirkwa:
@coderabbitai
Addressed in 2578653 — both `set_position` and `set_heading_true` accept `Option<f64>` and silently drop `None`, so the previous code armed the probe-satisfied flag even on a `value: null` delta. Now I read the Option locals first, pass them to the setter, and only call `mark_signalk_own_ship_nav_seen()` when those Options are `Some`. Tests still green (226 passing).

PR #264, opened 2026-05-26
fix(web): respect X-Forwarded-Proto and Host port for advertised WS URLs

stigdreyer:
## Motivation

When Mayara runs behind a reverse proxy that terminates TLS (e.g. Traefik at `https://halos.local:4434` forwarding to `http://localhost:6502`), the viewer's WebSocket connections fail because the server advertises the wrong URLs.

The `GET /signalk` and `GET /signalk/v2/api/vessels/self/radars` handlers were stripping the port from the `Host` header and replacing it with the internal `--port` value (6502), and deriving `ws`/`wss` solely from whether Mayara itself had TLS certs loaded — always `false` when TLS is terminated upstream. The viewer JS uses `radarInfo.streamUrl` and `radarInfo.spokeDataUrl` from the radars response in preference to `window.location`, so these advertised URLs must reflect the public-facing address.

## Solution

Introduce `derive_public_base()` in `web/mod.rs` used by both handlers:

- **Host:** uses the `Host` header as-is, preserving any external port already present (e.g. `:4434`); only appends the fallback port when no port is in the header (i.e. standard ports like 80/443)
- **Scheme:** checks `X-Forwarded-Proto` first, falls back to `state.tls`

Works correctly for:
- Direct access: `http://halos.local:6502` → `ws://halos.local:6502`
- Via Traefik TLS: `https://halos.local:4434` + `X-Forwarded-Proto: https` → `wss://halos.local:4434`
- Native TLS (cert/key flags): no proxy, `state.tls = true` → `wss://...`

No configuration changes required on the Mayara side — Traefik just needs to forward the original `Host` header (its default behavior).

## Summary

Adds derive_public_base() in web/mod.rs to derive externally-visible host, HTTP scheme, and WebSocket scheme from request headers and server TLS state, so advertised WebSocket URLs are correct behind TLS-terminating reverse proxies.

### Changes

**web/mod.rs**
- Adds pub(crate) fn derive_public_base(headers: &hyper::header::HeaderMap, tls: bool, fallback_port: u16) -> (String, &'static str, &'static str)
  - Chooses host by preferring X-Forwarded-Host (if present) and otherwise using Host, preserving any port present in the header.
  - Chooses scheme by preferring X-Forwarded-Proto (parses case-insensitively and takes the first comma-separated token) and falling back to the server TLS state.
  - Fixes port edge case: when X-Forwarded-Proto is present and the selected host has no port, do not append the internal fallback_port (let default ports apply); only append fallback_port for direct access scenarios when no forwarded proto is present.
  - Documents that X-Forwarded-* headers are trusted unconditionally.
- Replaces manual Host parsing and scheme selection in the endpoints handler with derive_public_base.
- Adds unit tests covering no headers, native TLS, Host with/without port, X-Forwarded-Proto variants (mixed-case, comma chains), X-Forwarded-Host precedence, and proxy-overrides of native TLS.

**signalk/v2.rs**
- get_radars uses derive_public_base(&headers, state.tls, state.args.port) to compute host and ws scheme.
- get_radar removes prior manual Host/port construction and uses a simplified debug statement.
- Removes unused imports (http::Uri, str::FromStr).

### Behavior

- Direct access: http → ws, https → wss based on server TLS state.
- Reverse proxy with TLS termination: X-Forwarded-Proto and X-Forwarded-Host are respected (wss produced when proxy indicates TLS, Host/port from proxy preserved).
- Native TLS: server TLS state is used when forwarded proto is absent.

### Tests and notes

- Adds 14 unit tests for derive_public_base covering header combinations and edge cases.
- Notes the trust assumption for X-Forwarded-* headers (proxies should forward original Host; proxies that don't must be configured accordingly).

[Review Change Stack](https://app.coderabbit.ai/change-stack/MarineYachtRadar/mayara-server/pull/264?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/bin/mayara-server/web/mod.rs
+++ b/src/bin/mayara-server/web/mod.rs
@@ -15,7 +15,6 @@ use std::{
     collections::HashMap,
     io,
     net::{IpAddr, Ipv6Addr, SocketAddr},
-    str::FromStr,
     sync::Arc,
 };
 use thiserror::Error;
@@ -298,24 +297,55 @@ async fn root_redirect() -> Redirect {
     Redirect::to("/gui/")
 }
 
+/// Derive the public-facing host, HTTP scheme, and WS scheme from request headers.
+///
+/// Respects `X-Forwarded-Proto` for reverse-proxy setups (e.g. Traefik TLS
+/// termination) and uses the `Host` header as-is so that external ports are
+/// preserved rather than replaced with the internal listen port.
+pub(crate) fn derive_public_base(
+    headers: &hyper::header::HeaderMap,
+    tls: bool,
+    fallback_port: u16,
+) -> (String, &'static str, &'static str) {
+    let host = headers
+        .get(axum::http::header::HOST)
+        .and_then(|v| v.to_str().ok())
+        .map(|h| {
+            // Append fallback port only when the Host header has none.
+            if h.contains(':') {
+                h.to_string()
+            } else {
+                format!("{}:{}", h, fallback_port)
+            }
+        })
+        .unwrap_or_else(|| format!("localhost:{}", fallback_port));
+
+    let forwarded_proto = headers
+        .get("x-forwarded-proto")
+        .and_then(|v| v.to_str().ok());
+    let (http_scheme, ws_scheme) = match forwarded_proto {
+        Some("https") => ("https", "wss"),
+        Some(_) => ("http", "ws"),
+        None => {
+            if tls {
+                ("https", "wss")
+            } else {
+                ("http", "ws")
+            }
+        }
+    };
+
+    (host, http_scheme, ws_scheme)
+}
+
 async fn quit_handler(State(state): State<Web>) -> &'static str {
     let _ = state.shutdown_tx.send(());
     "bye\n"
 }
 
 async fn endpoints(State(state): State<Web>, headers: hyper::header::HeaderMap) -> Response {
-    let host: String = match headers.get(axum::http::header::HOST) {
-        Some(host) => host.to_str().unwrap_or("localhost").to_string(),
-        None => "localhost".to_string(),
-    };
-    let host = format!(
-        "{}:{}",
-        match Uri::from_str(&host) {
-            Ok(uri) => uri.host().unwrap_or("localhost").to_string(),
-            Err(_) => "localhost".to_string(),
-        },
-        state.args.port
-    );
+    let (host, http_scheme, ws_scheme) =
+        derive_public_base(&headers, state.tls, state.args.port);
 
     let mut endpoints = Endpoints {
         endpoints: HashMap::new(),
@@ -323,11 +353,6 @@ async fn endpoints(State(state): State<Web>, headers: hyper::header::HeaderMap) 
             version: VERSION,
             id: PACKAGE,
         },
-    };
-    let (http_scheme, ws_scheme) = if state.tls {
-        ("https", "wss")
-    } else {
-        ("http", "ws")
     };
     endpoints.endpoints.insert(
         "v2".to_string(),
--- a/src/bin/mayara-server/web/mod.rs
+++ b/src/bin/mayara-server/web/mod.rs
@@ -302,40 +302,193 @@ async fn root_redirect() -> Redirect {
 /// Respects `X-Forwarded-Proto` for reverse-proxy setups (e.g. Traefik TLS
 /// termination) and uses the `Host` header as-is so that external ports are
 /// preserved rather than replaced with the internal listen port.
+///
+/// `X-Forwarded-Proto` is trusted unconditionally; ensure mayara is only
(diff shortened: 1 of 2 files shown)


dirkwa:
Please exit draft mode

dirkwa:
Thanks for digging into this, Stig — the analysis in the description is spot-on (host-header port stripping + scheme-from-`state.tls`-only), and `derive_public_base` consolidates three near-identical Host-parsing blocks nicely. A few questions and one bug I'd like to think through before merge:

**1. Port edge case when `X-Forwarded-Proto: https` is set but `Host` has no port.** Lines 314-320 always append `fallback_port` (= the internal `--port`, e.g. 6502) when `Host` has no `:`. So if a proxy sends `Host: halos.local` + `X-Forwarded-Proto: https` (the standard HTTPS-on-443 case), we end up advertising `wss://halos.local:6502`, which is wrong on both port and reachability. The Traefik example in the PR description happens to send `Host: halos.local:4434` so the bug is hidden there. Suggested fix: when `Host` has no port, omit the port entirely from the advertised URL — let the protocol default (443/80) apply. Only fall back to `fallback_port` when neither `Host` carries a port *nor* `X-Forwarded-Proto` is set (i.e. direct-access).

**2. `X-Forwarded-Host` / RFC 7239 `Forwarded`.** The PR relies on the assumption that the proxy preserves the original `Host` header. Traefik default and nginx with `proxy_set_header Host \$host` do this; some setups don't (they overwrite `Host` with the backend address and pass the original via `X-Forwarded-Host`). Worth a sentence in the doc-comment of `derive_public_base` saying "assumes `Host` is preserved by the proxy" — or, more robustly, check `X-Forwarded-Host` first and fall back to `Host`. Mirror of the `X-Forwarded-Proto` pattern you already added.

**3. Trust boundary.** Anyone reaching mayara directly can spoof these headers and influence the URLs we hand out. The blast radius is small (a malicious URL just makes someone's viewer fail to connect), but the convention is to either (a) only trust `X-Forwarded-*` when the request came from a configured trusted-proxy CIDR / from loopback, or (b) document the assumption. Not asking you to implement (a) for this PR; just want to flag it so the next person doesn't trip over it. A one-line doc-comment on the function would do.

**4. Alternative worth considering: drop the URLs entirely?** [`web/gui/control.js:1851-1867`](https://github.com/MarineYachtRadar/mayara-server/blob/main/web/gui/control.js#L1851-L1867) already falls back to `window.location.host` when `radarInfo.streamUrl` / `spokeDataUrl` is missing, and the protocol is derived from `window.location.protocol`. If the server returned path-only URLs (or omitted the fields), the client would always do the right thing without us having to guess at proxy state. The downside is that this is part of the public OpenAPI surface, so external consumers might depend on absolute URLs — your call. Mentioning in case the simpler path appeals.

**5. Tests.** `derive_public_base` is a pure function — a handful of \`#[test]\` cases would lock in the corner cases (no-headers, Host-with-port, Host-without-port + \`X-Forwarded-Proto: https\`, native TLS, malformed proto). The web/ module has no existing tests, so this would be the first; nothing fancy needed, just \`#[test]\` in \`mod.rs\`. Would catch (1) above.

**Smaller stuff:**
- Case-insensitive match on `X-Forwarded-Proto` would be safer — some proxies emit `HTTPS` or `Https`.
- `get_radar` (in `signalk/v2.rs`) no longer needs the `host` derivation at all post-refactor — the diff already reflects that. Nice cleanup.

Happy to chat through any of this; the core direction is right.

stigdreyer:
@dirkwa — addressed everything actionable from your review:

**Done:**
- **Port edge case** — \`fallback_port\` is no longer appended when \`X-Forwarded-Proto\` is present and \`Host\` has no port. A portless \`Host: halos.local\` + \`X-Forwarded-Proto: https\` now produces \`wss://halos.local\` instead of \`wss://halos.local:6502\`.
- **X-Forwarded-Host** — \`X-Forwarded-Host\` is now checked before \`Host\`, mirroring the \`X-Forwarded-Proto\` pattern. Proxies that preserve \`Host\` (Traefik default, standard nginx) are unaffected. Doc comment updated to explain the behaviour.
- **Trust boundary** — doc comment on \`derive_public_base\` notes that \`X-Forwarded-*\` headers are trusted unconditionally.
- **Tests** — 14 \`#[test]\` cases covering: no headers, native TLS, Host with/without port, \`X-Forwarded-Proto\` with https/http/mixed-case/comma-chained/unknown, \`X-Forwarded-Host\` precedence and fallback, and proxy overriding native TLS.
- **Smaller stuff** — case-insensitive proto match and comma-chain handling both done.

**Still open for discussion:**
- **Dropping absolute URLs** — happy to hear your take. The JS fallback to \`window.location\` already works and would sidestep all this proxy guessing, but it's a public API surface change.

dirkwa:
@coderabbitai review

Issue #156, opened 2026-05-02
Garmin Fantom Pro (dual-range) — works to control, but spoke WebSocket stays silent


inspirity:
### Mayara version or commit ID

Mayara <commit/version>, Debian 12

### Radar brand

Garmin

### Radar model

Fantom Pro (dual-range), IDs gar0200A / gar0200B

### PCAP file

_No response_

### Screenshot or video

_No response_

### Describe the issue

## Setup needed before mayara could talk to the radar                                                                                                     
  - Added a secondary IP in `172.16.0.0/12` to the radar interface
    (`ip addr add 172.16.3.150/12 dev eth1`) — without it, multicast                                                                                        
    discovery worked but unicast control did not.           
  - Issued `transmitChannel` per channel via curl before any `power=2` PUT;                                                                                 
    otherwise mayara errors `transmitChannel value 0 is lower than minimum
    value 1`.                                                                                                                                               
  - A Garmin chartplotter on the same unmanaged switch was required to wake                                                                                 
    the radar — Fantom Pro does not beacon standalone, and mayara appears
    to be passive for Garmin discovery.                                                                                                                     
                                                                                                                                                            
  ## Core problem — control vs. spokes are mutually exclusive
                                                                                                                                                            
  The Fantom Pro will only do one of the following at a time on this network,                                                                               
  not both:
                                                                                                                                                            
  **With the chartplotter on the switch:** radar wakes, spokes paint on the
  mayara overlay, but mayara cannot change any settings — `power`, `range`,
  `gain`, etc. are dispatched without error but the radar ignores them. The
  chartplotter appears to be asserting control and overriding mayara's PUTs.                                                                                
   
  **Without the chartplotter on the switch:** mayara's REST/control surface                                                                                 
  is fully responsive on Range A (capabilities, range, gain, sea, rain all
  work), but the radar doesn't wake on its own. Even after forcing transmit                                                                                 
  via curl and confirming spoke packets arrive on the interface, `/spokes`                                                                                  
  WS streamed 0 bytes over 8 s for `gar0200A`. SignalK stream WS works fine
  on the same server.                                                                                                                                       
                                                            
  Net effect: today there is no configuration where mayara is both the                                                                                      
  controller and the spoke source for this radar. Looks like a master-
  arbitration / spoke-publisher gap in the Garmin code path.
                                                                                                                                                            
  ## Range B never engages
[...]

PR #267, opened 2026-05-26
feat(garmin): send CDM V2 heartbeat to wake Fantom radars (#156)

keesverruijt:
## What changes for you

If you have a Garmin Fantom-class radar (Fantom, Fantom Pro, …) and you don't have a Garmin chartplotter on the same network, mayara previously couldn't see your radar at all — Fantoms stay silent until they spot a peer device (an MFD) broadcasting on the LAN. With this change, mayara plays that role itself, so a Fantom plugged directly into the mayara host (or sharing a switch with it) should now wake on its own, appear in the GUI, and start streaming.

No configuration change required — the heartbeat starts automatically when mayara launches. Legacy GMR HD/xHD radars (which use a hardwired Ethernet-cable pin for the same handshake) are unaffected, and replay/emulator modes don't send anything on the wire.

This is the first half of [#156](https://github.com/MarineYachtRadar/mayara-server/issues/156) — the discoverability problem. The dual-range/spoke-publisher arbitration issues described in the rest of that issue need separate work.

## Technical detail

Mayara now sends Garmin's `0x038e` "CDM V2 heartbeat" — 30 bytes, multicast to `239.254.2.2:50050`. Verified against the decompiled `cdm_build_and_send_v2_heartbeat` at `0x030879a4` in the GPSMAP 9000 firmware.

- **Cadence.** 1 s for the first 30 ticks (matches the MFD's startup ramp), then 5 s.
- **NIC re-enumeration.** Each tick re-walks 172.16/12 interfaces so a freshly configured NIC starts participating without a mayara restart.
- **No self-loop.** Multicast loopback is disabled on the send socket so the locator's receive path doesn't ingest mayara's own packets.
- **Group affinity.** `GarminLocator` carries an `Arc<AtomicU8>` shared with the sender. The receive path mirrors the observed radar's `syc_group_id` into the atomic, so within ~5 s of any real radar heartbeat mayara is sending in the same boat-local CDM group.
- **Lifecycle.** `garmin::new()` spawns the sender task; skipped under `--replay` and under `--features emulator` with `--emulator`. The sender relies on tokio runtime teardown for shutdown (no explicit `SubsystemHandle` integration yet).

Wire format we emit (30 bytes total):

```
GMN header: 8e 03 00 00 16 00 00 00    msg_id=0x038e, payload_len=22
Payload (22 bytes):
  +00 02            version_marker = 2
  +02 64 10         product_id = 0x1064  (GPSMAP 9000, "lionshark")
  +05 05            product_subtype = 5
  +06 NN            syc_group_id (mirrors observed radar)
  +07 01            constant
  +08 00            service_count = 0
  +0c 00 00 00 00   service-section padding (firmware 16-byte minimum)
  +10 01 04         tail tag: type=1, len=4
  +12 NN NN NN NN   u32 LE sequence
```

## Status & known follow-ups

- **Not yet verified against a live Fantom.** Expected to work end-to-end; please test if you have one without an MFD on the LAN.
- `service_count=0` may not be accepted by every firmware revision — if not, swap to advertising one fabricated service entry (the firmware accepts both shapes).
- `0x1064` product_id collides if the user actually has a GPSMAP 9000 on the same LAN; switching to `0xFFFE` is trivial if that turns out to matter.
- No graceful-shutdown integration on the sender task.

Refs #156

### DIFF
--- a/src/lib/brand/garmin/cdm_heartbeat.rs
+++ b/src/lib/brand/garmin/cdm_heartbeat.rs
@@ -1,0 +1,266 @@
+//! CDM "V2 heartbeat" sender — wakes a sleeping Fantom radar.
+//!
+//! Garmin radars (Fantom, Fantom Pro and likely xHD2/xHD3 as well) gate
+//! their report stream on observing a peer CDM device on the network.
+//! The MFD signals "I'm here" by multicasting `0x038e` to
+//! `239.254.2.2:50050` every 5 s (every 1 s for the first 30 sends
+//! after boot — see `cdm_build_and_send_v2_heartbeat` at
+//! `0x030879a4` in the MFD firmware decompilation).
+//!
+//! Legacy GMR HD/xHD radars achieve the same handshake via a hardwired
+//! pin in the Ethernet cable that an attached chartplotter energises;
+//! Fantom-class radars dropped that pin in favour of the Ethernet
+//! mechanism. Without anyone sending the heartbeat, the radar stays
+//! quiet and mayara never sees it.
+//!
+//! ## Wire format (30-byte UDP datagram)
+//!
+//! ```text
+//! GMN header (8 bytes):
+//!   8e 03 00 00       msg_id = 0x038e (LE)
+//!   16 00 00 00       payload_len = 22
+//!
+//! Payload (22 bytes):
+//!   +00 02            version_marker
+//!   +01 00            pad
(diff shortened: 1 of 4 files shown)


dirkwa:
@coderabbitai review

dirkwa:
I am limited too 🤣

PR #296, opened 2026-05-31
fix(gui): plot AIS using radar heading when SignalK true heading absent

dirkwa:
## Motivation

The AIS overlay only drew vessels when `this.trueHeading` (the SignalK true-heading subscription) was set. But the radar image already rotates fine on `this.lastHeading` (radar-derived heading). Behind a SignalK proxy the viewer commonly runs without a heading subscription, so `trueHeading` stays `null` and **every AIS vessel was silently skipped** — even with the sweep rotating and own-ship position known. Symptom: AIS vessels load (`aisVessels.size > 0`, `showAis: true`, own-ship set) but none render.

## Approach

Resolve an effective heading (radians) that prefers `trueHeading` and falls back to `lastHeading` (degrees → radians) — the same source preference the radar image uses — and gate/rotate the AIS draw on that. Vessels now plot whenever the display has any heading reference.

## Tested

- `cargo build --release` and `cargo test` pass (15 AIS-store/subscription tests green; change is GUI JS).
- Verified the new gate against a real failing session (`trueHeading: null`, `lastHeading: -46.43°`, own-ship set, 45 vessels): previously skipped, now draws. SignalK-heading-present and no-own-ship cases unchanged; the genuine no-heading case still skips.

Adds a private helper method `#effectiveHeadingRad()` that returns an overlay heading in radians by preferring the Signal K true heading and falling back to the radar-derived heading (converting from degrees to radians), or returns null when neither is available. This method mirrors the source preference the radar image already uses for rotation.

Updates AIS vessel rendering in `#drawAisVessel()` to use this effective heading instead of requiring the true heading to be present. Replaces the check `if (this.trueHeading === null) return;` with a call to `#effectiveHeadingRad()`, allowing AIS vessels to render whenever any heading reference is available. This resolves the case where a viewer runs behind a Signal K proxy without a heading subscription—the sweep rotates fine on radar-derived heading, but AIS vessels were previously hidden entirely despite having position and vessel data available.

### DIFF
--- a/web/gui/ppi.js
+++ b/web/gui/ppi.js
@@ -301,6 +301,17 @@ class PPI {
     return this.trueHeading;
   }
 
+  // Heading in radians for overlay placement, preferring the Signal K
+  // true heading and falling back to the radar-derived heading
+  // (`lastHeading`, stored in degrees). Mirrors the source preference the
+  // radar image already uses so AIS plots whenever the sweep can rotate.
+  // Returns null when no heading source is available.
+  #effectiveHeadingRad() {
+    if (this.trueHeading !== null) return this.trueHeading;
+    if (this.lastHeading !== null) return (this.lastHeading * Math.PI) / 180;
+    return null;
+  }
+
   getHeadingMode() {
     return this.headingMode;
   }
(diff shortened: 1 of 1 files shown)


dirkwa:
@coderabbitai Good catch — the finding is valid. `#drawTarget` (line 1180/1193/1250) has the same overly-strict `trueHeading`-only gate as the AIS path, so ARPA targets also won't render when only the radar-derived heading is available, and the fix (use `#effectiveHeadingRad()` for the draw + velocity-vector math) is correct. I also confirmed the MARPA acquire path keeps its hard `trueHeading` requirement — but that check already lives in the separate click handler (`acquireTargetMode`, ~line 2110: "MARPA requires true bearing"), so it's untouched by a draw-path change.

However, this repo enforces one logical change per PR (separate changelog entries → separate PRs). ARPA-target rendering is a distinct change from this AIS PR and would be its own changelog line, so I'm keeping #296 AIS-only and opening a separate PR for the ARPA `#drawTarget` heading fallback rather than bundling it here.

PR #313, opened 2026-06-01
fix(gui): show own-ship nav and heading UI without a heading subscription

dirkwa:
## Why

Behind the Signal K proxy the viewer detects "standalone" mode (mayara answers `/signalk` through the proxy), and in that mode it skipped the `vessels.self` subscription for `navigation.headingTrue` / `navigation.position`. Own-ship position and heading then came only from radar spokes, which are produced while transmitting — so in standby the viewer showed no position, no heading, the H-Up toggle stayed disabled, and the AIS overlay had no own-ship reference to plot against.

A secondary issue: when only the radar-derived heading is available, AIS/ARPA targets jumped frame to frame. The per-spoke `bearing` field is quantized, and in head-up mode the overlay subtracts that heading while the spoke raster does not, so the jitter moved every target.

## How

- Always open the `vessels.self` nav subscription. mayara serves own-ship position and heading from its upstream regardless of mode or radar transmit state, so the viewer no longer depends on spokes for them.
- Enable the H-Up toggle and AIS lozenge from the radar-derived heading when no Signal K heading is published.
- Smooth the radar-derived heading on its unit vector (circular-mean safe across the wrap) to stop the AIS/ARPA jump, and fall back to it in the position box instead of showing a misleading `0`.

## Tested

Verified live against a DRS4D-NXT behind a Signal K server, both the proxied (`/plugins/<id>/gui/`) and direct (`:6502`) URLs: own-ship position and heading display in standby, the H-Up toggle is clickable, and AIS targets are stable. `cargo test` passes.

## Summary

The GUI now reliably displays own-ship position and heading in all modes, including when the viewer is behind a Signal K proxy or when the radar is in standby, and stabilizes AIS target placement.

### web/gui/ppi.js

Introduces exponential smoothing (circular mean via unit-vector smoothing) for radar-derived heading to eliminate jitter. A new `HEADING_SMOOTH_ALPHA` constant sets the smoothing factor to 0.1. The `drawSpoke()` method now:
- Computes spoke-based heading from bearing and angle
- Incrementally smooths the heading using sine/cosine components (`headingSin`/`headingCos`) to handle the 360°→0° wraparound safely
- Derives `lastHeading` from the smoothed components using `atan2`
- Clears the smoothed state when spoke heading inputs are missing

### web/gui/viewer.js

Refactors heading handling to always subscribe to `vessels.self` navigation streams (position and heading) and to enable heading-dependent UI based on either Signal K or radar-derived heading. Key changes:

- Adds `haveSignalKHeading` flag to track arrival of the first real heading delta from Signal K
- `subscribeToHeading()` now unconditionally opens the `vessels.self` subscription for `navigation.headingTrue` and `navigation.position`, ensuring own-ship data are available regardless of mode or radar TX state
- Introduces `positionBoxHeadingDeg()` helper that prefers Signal K true heading once a delta arrives, otherwise falls back to the radar-derived heading
- Spoke processing conditionally calls `onHeadingReceived()` once when radar spokes provide a usable heading but the Signal K socket hasn't yet fired, gated to avoid re-triggering during brief radar gaps
- `onHeadingReceived()` enables the H-Up toggle and AIS lozenge when any heading becomes available
- Heading display in the position box stays in sync across periodic refreshes and position updates

### DIFF
--- a/web/gui/ppi.js
+++ b/web/gui/ppi.js
@@ -9,6 +9,17 @@ const RANGE_SCALE = 0.9;
 
 const NAUTICAL_MILE = 1852.0;
 
+// Exponential-smoothing factor for the radar-derived heading. The bearing
+// field on each spoke is quantized, so the per-spoke `bearing - angle` heading
+// wobbles by a fraction of a degree frame to frame even with a steady boat.
+// In head-up mode the AIS/ARPA overlay subtracts this heading to place
+// geographic targets bow-up (the spoke raster is drawn bow-up and does NOT
+// subtract it), so the wobble shows up as targets jumping. Smoothing the
+// heading on its unit vector (circular-mean safe across the 360->0 wrap) kills
+// the jitter while staying responsive: at typical spoke rates the time
+// constant is well under the timescale of a real turn.
+const HEADING_SMOOTH_ALPHA = 0.1;
+
 function divides_near(a, b) {
   let remainder = a % b;
   return remainder <= 1.0 || remainder >= b - 1;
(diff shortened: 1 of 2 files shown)


dirkwa:
@coderabbitai review

PR #319, opened 2026-06-01
ci: add opt-in per-PR Docker test image build

dirkwa:
Reviewers sometimes want to pull and run the image from a specific PR without waiting for a release. There was no way to get a Docker image for an in-flight PR — only `main` and tags build images.

This adds an opt-in workflow that builds a multi-arch image for a single PR, tagged `pr<NUMBER>` (e.g. `pr123`):

- A maintainer adds the `build-image` label to a PR to enable it; while the label is present the image rebuilds on every push, and removing the label stops it. Since only maintainers can label, this also gates builds from forks.
- Reuses the existing cross-build → artifact → multi-arch assembly from `docker.yml` and the same `docker/Dockerfile.ci`. Builds `linux/amd64` + `linux/arm64`.
- Pushes `pr<N>` (moving) and `pr<N>-<sha>` (immutable) tags to GHCR, then posts/updates a single PR comment with the `docker pull` command.

Requires a `build-image` label to exist in the repo.

### DIFF
--- a/.github/workflows/docker-pr.yml
+++ b/.github/workflows/docker-pr.yml
@@ -1,0 +1,154 @@
+name: Docker PR test image
+
+# Builds an opt-in Docker test image for a single pull request, tagged `pr<NUMBER>`
+# (e.g. `pr123`). Maintainers enable it by adding the `build-image` label to a PR;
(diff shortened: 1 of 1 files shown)


keesverruijt:
Great!

PR #325, opened 2026-06-01
feat(garmin): skip spoke decode in idle mode to free CPU at idle

dirkwa:
## Why

PR #284 found Furuno radars emit spokes continuously even in Standby, wasting CPU decoding frames nobody watches. Garmin's data path decodes every spoke with no idle gating, so the same waste likely applies.

## How

Reuses the brand-agnostic idle gate from #321:
- Gate both spoke paths — `process_data` (xHD) and `process_hd_spoke` (HD, which arrives on the report port) — on a `both_ranges_idle()` check.
- Refresh the per-range idle flags from `process_report` (reports arrive regularly, so no separate timer is needed).
- Wake on a non-Standby power transition in both `process_transmit_state` (xHD) and `process_hd_status` (HD). The web layer already wakes on spoke-WS subscribe and control PUT.

The spoke decoder is stateless per frame, so no delta-buffer reset is needed. In dual-range mode the radar stays non-idle unless **both** ranges report Standby — the conservative, correct default.

**Safe by design:** if a Garmin radar stops emitting spokes in Standby, the gate never triggers and is a harmless no-op.

## Testing

Builds clean, `cargo test` passes (268 + 15). I do not have a Garmin radar to measure idle CPU. **Testers with a Garmin (xHD/HD/Fantom) radar:** please report mayara's CPU with the radar in Standby and no GUI/WS client connected, before and after this change, and confirm the PPI still appears instantly when you open the viewer or start transmitting.

### DIFF
--- a/src/lib/brand/garmin/report.rs
+++ b/src/lib/brand/garmin/report.rs
@@ -418,6 +418,15 @@ impl GarminReportReceiver {
             bail!("Report too short: {} bytes", data.len());
         }
 
+        // Reports arrive regularly, so refresh idle state here rather than on a
+        // separate timer. The web layer wakes us on spoke subscribe, so a late
+        // entry into idle only wastes a little CPU briefly — never blanks a
+        // viewer.
+        self.common.info.refresh_idle_flag();
+        if let Some(ref cb) = self.common_b {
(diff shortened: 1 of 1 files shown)


keesverruijt:

PR #327, opened 2026-06-01
feat(gui): surface upstream nav/AIS status as a banner

dirkwa:
## Summary

Stacked on #326 (base is that branch, so this diff is just the GUI change).

When mayara's upstream SignalK connection has no own-ship position, or no approved device token (so the authenticated AIS REST seed is skipped), the radar viewer showed an empty overlay with no explanation — the exact symptom a user hit on an auth-enabled port-80 server with a pending access request.

The viewer now polls the `nav` block #326 adds to `/signalk` and shows a non-blocking amber banner:

- no own-ship position yet → "No SignalK navigation yet — own-ship position & AIS unavailable."
- position flowing but token pending/denied and no AIS → "Waiting for SignalK token approval — open Security → Access Requests to enable AIS."

The banner auto-clears once position/AIS arrive (it polls every 5s while a banner is up, 30s when healthy), and stays silent for non-SignalK nav sources (`static`/`nmea0183`/`udp`) so standalone setups never see a false warning. All URLs go through `apiBase()` so it works both proxied and direct.

## Tested

- `node --check` on viewer.js; `cargo build` embeds the updated GUI (confirmed the banner code is served at `/gui/viewer.js`).
- Decision logic (`navStatusMessage`) unit-checked across tcp/ws/static/nmea0183 + healthy/seeding states — all correct, including no false banner in standalone mode.
- Ran the built binary in `tcp:` mode; `/signalk` reports `nav:{transport:tcp,token_present:false,have_position:false}`, which maps to the "No SignalK navigation yet" banner.

### DIFF
--- a/web/gui/layout.css
+++ b/web/gui/layout.css
@@ -133,6 +133,30 @@ div.myr_ppi {
   text-align: center;
 }
 
+/* Non-blocking banner across the top of the PPI, shown when mayara reports
+   its upstream SignalK nav/AIS is unavailable (e.g. an unapproved device
+   token). Auto-clears once the condition resolves. Amber so it reads as a
+   notice, not a hard error, over the green radar aesthetic. */
+.myr_status_banner {
+  position: absolute;
+  top: 12px;
+  left: 50%;
+  transform: translateX(-50%);
+  max-width: 90%;
+  padding: 8px 14px;
+  background: rgba(60, 40, 0, 0.85);
+  color: #ffcc44;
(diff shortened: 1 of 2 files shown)


dirkwa:
Note: GitHub auto-marked this as merged when #326's branch was fast-forwarded to include this commit. The GUI banner is **not** on `main` yet — it ships as part of #326, which now contains both the `/signalk` status field and this banner as one logical change. Track #326.

PR #334, opened 2026-06-03
feat(raymarine): surface radars that announce no data stream

dirkwa:
## Summary

- A Quantum reachable only through a Raymarine MFD acting as a WiFi access point announces an unspecified report address (`0.0.0.0`) in its beacon — there is no spoke multicast group to join, so the report socket fails to bind (EINVAL) and no radar image ever appears, with nothing on screen to explain why.
- Detect the unspecified report address at discovery: keep the radar listed (so the operator can see it), warn once with a remedy, and report it on the `/signalk` endpoint as `dataStreamMissing`. The GUI surfaces this in the existing yellow status banner (the same one used for token/nav issues), auto-clearing once a real stream appears.
- The reliable fix for the user is to connect the radar via wired Ethernet; this change makes the failure visible instead of silent.

## Tested

- New unit test `decode_quantum_with_no_data_stream` using a real captured Quantum2 beacon (link_id 0xD681C8C3 behind an Axiom): asserts the radar is created and commandable but `has_data_stream()` is false.
- Full `cargo test` green; `cargo build` clean.

## Overview

Detects and surfaces Quantum radars that announce an unspecified report address (0.0.0.0) in their beacon. These radars remain listed and commandable but are flagged as missing a data stream. The locator emits a one-time warning per radar, the condition is reported on the /signalk endpoint, and the GUI shows the existing yellow status banner until a real stream appears.

### Radar health detection
- Adds `pub fn has_data_stream(&self) -> bool` on `RadarInfo` that derives health from the radar’s spoke-data endpoint (spoke_data_addr) — returning true only when the IP is not unspecified and the port is non-zero.
- Adds `pub fn get_all(&self) -> Vec<RadarInfo>` on `SharedRadars` to return all discovered radars (including those without active ranges).

### Raymarine beacon processing and warnings
- Raymarine locator tracks warned radars with a `HashSet<LinkId>` and logs a one-time “no data stream” warning when a Quantum beacon advertises an unspecified report IP (0.0.0.0). Radars are still created even when the report/spoke address is unspecified, allowing visibility/commandability while marking them as missing the data stream.

### Report receiver behavior
- `RaymarineReportReceiver::run` detects an unspecified report address (0.0.0.0) and exits immediately (Ok(())) instead of entering the socket-retry busy loop, avoiding a spinning/unbindable receiver when no data stream exists.

### SignalK endpoint
- Introduces a `RadarHealth` structure into the `/signalk` JSON response and adds `radars: Vec<RadarHealth>` to the Endpoints response. The handler populates this list by iterating all radars and setting `data_stream_missing: true` for radars where `has_data_stream()` is false.

## Frontend Changes

- `viewer.js` adds `radarStatusMessage(radars)` and updates status polling so the existing yellow banner displays an explanatory message when a radar is reachable but missing its data stream; navigation/connectivity messages retain higher precedence. The banner auto-clears once a real stream appears.

## Testing

- Adds unit test `decode_quantum_with_no_data_stream` using a captured Quantum2 beacon to assert the radar is created and commandable while `has_data_stream()` returns false. Cargo tests and build succeed.

## Notes / Scope
- The change intentionally does not implement an in-place 0.0.0.0 → real-address identity transition. Address-independent identity and receiver rebind are out of scope; a radar that switches to a real multicast address is discovered as a new registry entry and the stale ray0000 entry clears on restart. The primary goal is to make the failure visible and avoid a busy-looping receiver; the recommended user remedy is to connect the radar via wired Ethernet.

### DIFF
--- a/src/bin/mayara-server/web/mod.rs
+++ b/src/bin/mayara-server/web/mod.rs
@@ -269,6 +269,23 @@ struct Endpoints {
     /// ignored by standard clients.
     #[schema(value_type = Object)]
     nav: mayara::navdata::NavStatus,
+    /// mayara-specific per-radar health, so a GUI can surface a radar that was
+    /// discovered but announced no data stream (e.g. a Quantum behind a WiFi
+    /// access point). Not part of the Signal K spec; ignored by standard
+    /// clients.
+    radars: Vec<RadarHealth>,
+}
+
+/// Health summary for a single discovered radar, surfaced on the `/signalk`
+/// endpoint alongside `nav` for the GUI status banner.
+#[derive(Serialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+struct RadarHealth {
+    /// Radar id (e.g. `ray0000`).
+    id: String,
+    /// False when the radar announced no usable spoke/report address, so no
+    /// radar image can appear until the network path is fixed.
+    data_stream_missing: bool,
 }
 
 #[derive(Serialize, ToSchema)]
@@ -543,6 +560,19 @@ async fn endpoints(State(state): State<Web>, headers: hyper::header::HeaderMap) 
     let (host, http_scheme, ws_scheme) =
         derive_public_base(&headers, state.tls, state.args.port);
 
+    // Surface only radars that need attention; a healthy fleet sends an empty
+    // list so the GUI shows nothing.
+    let radars = state
+        .radars
+        .get_all()
+        .iter()
+        .filter(|info| !info.has_data_stream())
+        .map(|info| RadarHealth {
+            id: info.key(),
+            data_stream_missing: true,
+        })
+        .collect();
+
     let mut endpoints = Endpoints {
         endpoints: HashMap::new(),
         server: Server {
@@ -550,6 +580,7 @@ async fn endpoints(State(state): State<Web>, headers: hyper::header::HeaderMap) 
             id: PACKAGE,
         },
         nav: mayara::navdata::nav_status(&state.args),
+        radars,
     };
     endpoints.endpoints.insert(
         "v2".to_string(),
--- a/src/lib/brand/raymarine/mod.rs
+++ b/src/lib/brand/raymarine/mod.rs
@@ -1,6 +1,6 @@
 use anyhow::{Error, bail};
 use serde::Deserialize;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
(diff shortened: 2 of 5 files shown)


dirkwa:
@coderabbitai review

dirkwa:
@coderabbitai Thanks — addressed in 4d1c99a.

1. **`has_data_stream()` → `spoke_data_addr`** (Major, inline): done. The helper now checks `spoke_data_addr` (and a non-zero port) instead of `report_addr`, so the flag reflects the actual spoke endpoint and generalises across brands. For Raymarine the two are the same value here, so behaviour for this case is unchanged.

2. **`0.0.0.0 → real address` transition** (Major, outside-diff): not handled in this PR, by design. A Raymarine radar's key is derived from its address (`addr = data.report`), so the unspecified state keys as `ray0000` while a real multicast group keys as `rayXXXX` — they are *different* registry entries, not an in-place update. A correct in-place transition requires making radar identity address-independent (e.g. keying by `link_id`) plus receiver rebind, which touches identity/persistence well beyond this "fail visibly" change and belongs in its own PR (per AGENTS.md one-logical-change).

   In the meantime: the stale `ray0000` receiver no longer busy-loops on the unbindable `0.0.0.0` socket — it now exits immediately (report.rs). A rewired radar re-discovers cleanly as a working `rayXXXX` entry; the stale entry clears on the next restart. The intended user remedy (wired Ethernet) yields a fresh discovery with a real address, so no transition occurs in practice.

keesverruijt:
Closing as stale: superseded by #348.

This PR was a UX workaround for a failure mode the codebase couldn't recover from — a Quantum behind an MFD WiFi AP advertises `report=0.0.0.0:0`, the report socket couldn't bind, and no image ever appeared. The remedy was to surface the failure as a yellow banner so the operator knew to wire the radar over Ethernet.

#348 ([merged in d0bd72f](https://github.com/MarineYachtRadar/mayara-server/commit/d0bd72f)) actually fixes the underlying topology: when the beacon's report address is unspecified, mayara now falls back to a unicast listen on the command port and shares a single connected socket between the command sender and the report receiver. The radar produces an image; the trigger condition `dataStreamMissing` was watching for no longer fires.

If a genuinely unrecoverable "discovered but never streams" failure mode shows up later, the health-indicator pattern from this PR (`has_data_stream()` + Signal K flag + GUI banner) is worth reviving — but tied to a real symptom we can reproduce, not the WiFi-AP case which is now functional.

PR #346, opened 2026-06-08
feat(navico): surface radar errors and a Fault power state

### LINKED ISSUE #345: feat(navico): surface radar error reports (0xC611) and a Fault power state
## Problem

mayara already receives the Navico active-error report `0xC611` but mishandles it. In `src/lib/brand/navico/report.rs` (`process_report`):

```rust
0x11 => {
    if data.len() != 3 || data[2] != 0 {
        bail!("Strange content of report 0x0a 0xc6: {:02X?}", data);
    }
    // this is just a response to the MFD sending 0x0a 0xc2,
    // not sure what purpose it serves.
}
```

That `data[2]` it asserts is `0` is the **active-error count**. So:
- a **healthy** radar (count 0) is ignored with an incorrect comment, and
- a **faulting** radar (count ≥ 1) hits `bail!` — mayara discards the one fault report it already sees.

## Background (RE)

Reverse-engineering of the Navico NRP protocol (ZEUS SRX 2.5.98 `libAndroidRadarService`) documents the error channel. See `research/navico/radar-error-reporting.md`:

- **`0xC611`** active-error list: `[u16 op][u8 count][count × u32 errorType]`.
- **`0xC610`** single error (not seen on the report channel mayara uses).
- `errorType` is `NRP::eRadarErrorType` — a ~31-entry enum (`0x00001`, `0x10001`–`0x1001E`): `MotorNotRunning`, `CommsNotActive`, `PAHardOverheat`, `PSUVoltage`, `DopplerFail`, … The enum grows by appending higher values in newer firmware.

Every Navico capture we have shows `0xC611 count=0` (healthy); no non-zero report has been captured, so per-error `param` semantics are unknown. Decode the **framing** now; pass raw codes through.

## Proposed change (one PR)

1. Replace the `0x11 0xc6` handler with a real `0xC611` parser: read `count`, then `count × u32` error type IDs.
2. Map each code to an **optional** human name (we don't know all strings; unknown codes pass through with `name = None`).
3. Surface the active-error set on the radar state exposed to clients.
4. Add a single new `Power` state: **`Fault`** (a failed self-test is still a fault — one state, not a separate variant). Driven by the presence of active errors.

Self-test/diagnostics telemetry (Navico SRT `0xC82x`) is **out of scope** — no genuine SRT packet appears in any capture; defer until a real capture exists.

Testing without a broken radar: extend the existing `--fake-errors` mechanism to inject a fake active-error entry.

keesverruijt:
## Why

mayara already receives the Navico `0xC611` report — the radar's **active-error list** — but the handler asserted its count byte was always `0` and `bail!`ed otherwise, discarding the one fault report it already sees. A faulting radar was therefore invisible (and noisy in the log).

Background: `research/navico/radar-error-reporting.md` (RE of the NRP protocol, ZEUS SRX `libAndroidRadarService`).

## How

- Parse `0xC611` as `[count][count × u32 errorType]` (`parse_radar_errors`).
- Map each NRP `eRadarErrorType` to an **optional** name — unknown/newer codes pass through with no name, surfaced by their raw code.
- Raise/clear a Signal K `notifications.radar.<id>.error.<code>` per error, mirroring the existing guard-zone notification.
- Add a single `Power::Fault` state (a failed self-test is still a fault), shown while any error is active and lifted when they clear.

Out of scope: Navico SRT `0xC82x` self-test/diagnostics telemetry — no genuine SRT packet appears in any capture, so it's deferred until a real capture exists.

## Test plan

- [x] `cargo test` — 283 pass, incl. new unit tests for `0xC611` parsing (empty/active/malformed) and the code→name lookup.

closes #345

Implements correct parsing and surfacing of Navico NRP active-error report 0xC611, which the prior handler effectively discarded when the report includes active errors. The receiver now extracts the active error set, publishes Signal K notifications per error code, and forces the radar power state into a new Fault power state while any error is active.

## Changes in `src/lib/brand/navico/report.rs`

`NavicoReportReceiver` tracks `active_errors: BTreeSet<u32>` (the current active error codes) and `reported_power: Power` (the last power value reported by the radar). During normal status handling (`set_status`), it:
- maps the radar status byte to a `Power`
- stores it in `reported_power`
- publishes `ControlId::Power` as `Power::Fault` whenever `active_errors` is non-empty; otherwise it publishes `reported_power`

When a 0xC611 packet is received (identified by `data[0] == 0x11` and the radar report subtype already gating `data[1] == 0xc6`), `process_report` routes it to `process_radar_errors()`, which:
- parses the payload via `parse_radar_errors()` using the `[0x11 0xc6][u8 count][count × u32 LE]` layout
- diffs the newly parsed set against the previous `active_errors`
- emits Signal K notifications for newly-active vs cleared codes via `controls.send_radar_error_notification()`
- updates `active_errors` and refreshes the effective power state via `publish_power()`

`parse_radar_errors()` validates the expected structure and length, and `error_name()` maps known NRP error codes to human-readable names while returning `None` for unknown/newer codes so raw codes can still be surfaced.

Unit tests cover 0xC611 parsing for empty lists, active/multi-error lists, malformed/length-mismatch cases, and known-vs-unknown code-to-name mapping.

## Changes in `src/lib/radar/mod.rs`

`Power` gains a new `Fault` variant and derives `Clone`/`Copy`. `Power::from_value` accepts fault values in both numeric and string forms (including `4` and `"fault"`-style inputs).

## Changes in `src/lib/radar/settings.rs`

`ControlId::Power` now accepts `"Fault"`. A new `Controls::send_radar_error_notification(code, name, active)` method emits:
- `Alarm` notifications with visual+sound methods and an error message derived from `name` (or hex code if `name` is unknown) when `active == true`
- `Normal` notifications clearing the prior alarm (with a cleared message and no methods) when `active == false`
- broadcasts the notification deltas via the Signal K client transaction

## Changes in `src/lib/brand/emulator/report.rs`

The emulator treats both `Power::Preparing` and `Power::Fault` as “no-op” power updates in `EmulatorReportReceiver::handle_control_update`, matching the previous behavior for `Preparing`.

### DIFF
--- a/src/lib/brand/emulator/report.rs
+++ b/src/lib/brand/emulator/report.rs
@@ -150,7 +150,7 @@ impl EmulatorReportReceiver {
                             self.common
                                 .set_value(&ControlId::Power, Power::Off as i32 as f64);
                         }
-                        Ok(Power::Preparing) | Err(_) => {}
+                        Ok(Power::Preparing | Power::Fault) | Err(_) => {}
                     }
                 }
             }
--- a/src/lib/brand/navico/report.rs
+++ b/src/lib/brand/navico/report.rs
@@ -2,6 +2,7 @@ use anyhow::{Error, bail};
 use num_traits::FromPrimitive;
 use serde::Deserialize;
 use std::cmp::min;
+use std::collections::BTreeSet;
 use std::io;
 use std::mem::transmute;
 use std::net::SocketAddr;
@@ -185,6 +186,13 @@ pub(crate) struct NavicoReportReceiver {
     reported_unknown: [bool; 256],
     reported_setup_ext_oversize: bool,
     has_use_mode_from_ext: bool,
+
+    /// Active radar error codes (NRP eRadarErrorType) from the 0xC611 report.
+    active_errors: BTreeSet<u32>,
+    /// Power state last reported via the 0xC4xx state-mode report, kept so the
+    /// fault override (active errors -> [`Power::Fault`]) can be lifted once the
+    /// errors clear without waiting for the next state-mode report.
+    reported_power: Power,
 
     // For data (spokes)
     data_buf: Vec<u8>,
@@ -416,6 +424,8 @@ impl NavicoReportReceiver {
             reported_unknown: [false; 256],
             reported_setup_ext_oversize: false,
             has_use_mode_from_ext: false,
+            active_errors: BTreeSet::new(),
+            reported_power: Power::Standby,
             data_buf: Vec::with_capacity(size_of::<RadarFramePkt>()),
             data_socket: None,
             doppler: DopplerMode::None,
@@ -762,11 +772,7 @@ impl NavicoReportReceiver {
             if data[1] == 0xc6 {
                 match data[0] {
                     0x11 => {
-                        if data.len() != 3 || data[2] != 0 {
-                            bail!("Strange content of report 0x0a 0xc6: {:02X?}", data);
-                        }
-                        // this is just a response to the MFD sending 0x0a 0xc2,
-                        // not sure what purpose it serves.
+                        return self.process_radar_errors();
                     }
                     _ => {
                         log::trace!("Unknown report 0x{:02x} 0xc6: {:02X?}", data[0], data);
@@ -828,7 +834,7 @@ impl NavicoReportReceiver {
     }
 
     fn set_status(&mut self, status: u8) -> Result<(), Error> {
-        let status = match status {
+        let power = match status {
             0 => Power::Off,
             1 => Power::Standby,
             2 => Power::Transmit,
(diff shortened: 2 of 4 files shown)


keesverruijt:
## CodeRabbit follow-ups

Two findings addressed in the rebased commit:

- `send_radar_error_notification` is now `pub(crate)` — matches the minimal-visibility rule from AGENTS.md.
- Added unit tests for the new `Power::Fault` parse branches in `Power::from_value` (int 4, float 4.0, `"fault"` / `"Fault"` / `"4"`, plus negative cases for `"faulty"`, `5`, `-1`).

Two findings deliberately skipped:

- **Emulator `Power::Fault` no-op test**: the new arm is literally `Ok(Power::Preparing | Power::Fault) | Err(_) => {}` — a syntactic no-op with no behavior to assert. Exercising it would mean constructing a full `EmulatorReportReceiver` (with `CommonRadar`, `EmulatorWorld`, sockets, etc.) just to verify "nothing happens". Cost-to-value is poor.
- **`settings.rs:996-1005` fractional `valid_values` check**: pre-existing code added by PR #358, not part of this PR's diff. Worth fixing, but out of scope for this PR per "one logical change per PR". Happy to file a follow-up issue if useful.

PR #348, opened 2026-06-08
fix(raymarine): listen unicast when Quantum announces no report multicast

dirkwa:
## What changes for you

If you have a Raymarine Quantum radar **and** you connect to it through an Axiom MFD that's acting as the WiFi access point — i.e., your mayara host joins the Axiom's WiFi network rather than the radar's own — mayara should now successfully discover the radar, populate its controls, and render the live image. Before this change, that setup failed in one of three ways depending on how far you got: the radar might not be discovered at all, it might appear in mayara but never produce an image, or commands sent from the GUI would silently get nowhere.

The fix is automatic — no new settings or flags. Direct-to-radar WiFi (where mayara joins the radar's own AP) and all other brands continue to behave exactly as before.

## Technical detail

A Quantum behind an MFD-as-WiFi-AP advertises `0.0.0.0:0` for its report group and streams reports/spokes back unicast to whichever host sent it a command. Getting that to work surfaced three sequential bugs, each found from real-hardware captures:

1. **Discovery / address.** The unspecified report address caused the receiver to try joining multicast `0.0.0.0:0`, which fails — no data arrives, the radar stays hidden. Now falls back to a unicast listen on the command port.
2. **Socket collision.** With (1) fixed, the report listener and the command sender both opened sockets on the same `host:port`; the connected command socket won delivery of the radar's replies, starving the listen socket. Share one connected `Arc<UdpSocket>` between the report loop (recv) and the command sender (send) so commands and replies travel on one source port.
3. **Fatal heartbeat send.** The shared connected socket means a heartbeat `send` can fail with a routing error (seen in a container where multicast binds also fail with `EADDRNOTAVAIL`). The report loop propagated that via `?`, dropping the report socket and respinning once per second forever, so no heartbeat ever left and the radar never streamed. Heartbeat-send failures are now non-fatal: log, advance the deadline, and keep listening.

The unicast path is gated on `report_addr` being non-multicast, so the standard multicast topology and every other brand are unchanged. The connected-socket bind log is at `debug` so `-v` shows the unicast path without trace-level noise.

## Tested

- Real hardware: Raymarine Quantum behind an Axiom MFD as WiFi AP (Raspberry Pi running this build via the Signal K container plugin, `--allow-wifi`). Radar discovered, controls populate, **radar echoes render** in the GUI.
- Unit test `quantum_behind_mfd_ap_uses_unicast_report_stream` (real beacons) asserts the unspecified report address falls back to `0.0.0.0:<command-port>`.
- `cargo test --lib` clean (310 tests pass, rebased onto current main).
- `cargo fmt --check` and `cargo clippy --no-deps --all-targets -- -D warnings` clean.
- Builds verified: default, `--features pcap-replay`, and each brand alone.

### DIFF
--- a/src/lib/brand/raymarine/command.rs
+++ b/src/lib/brand/raymarine/command.rs
@@ -17,6 +17,11 @@ pub(crate) struct Command {
     info: RadarInfo,
     model: BaseModel,
     sock: Option<Arc<UdpSocket>>,
+    /// Whether this sender may open its own command socket. False in the
+    /// unicast-stream topology, where it must share the report socket — see
+    /// `disable_own_socket`. Opening its own socket there would bind the same
+    /// host:port as the report socket and steal the radar's replies.
+    own_socket_allowed: bool,
 }
 
 impl Command {
@@ -26,15 +31,23 @@ impl Command {
             info,
             model,
             sock: None,
+            own_socket_allowed: true,
         }
     }
 
     /// Use a caller-provided, already-connected socket for sending instead
     /// of opening our own. Used by the unicast-stream topology so commands
     /// and the radar's replies share one socket (and one source port).
-    pub(crate) fn with_shared_socket(mut self, sock: Arc<UdpSocket>) -> Self {
+    pub(crate) fn set_shared_socket(&mut self, sock: Arc<UdpSocket>) {
         self.sock = Some(sock);
-        self
+        self.own_socket_allowed = false;
+    }
+
+    /// Forbid opening an own command socket while no shared socket is set
+    /// yet (unicast topology, socket creation pending a retry). Sends are
+    /// dropped until `set_shared_socket` supplies the shared socket.
+    pub(crate) fn disable_own_socket(&mut self) {
+        self.own_socket_allowed = false;
     }
 
     pub(crate) fn set_ranges(&mut self, ranges: Ranges) {
@@ -69,6 +82,12 @@ impl Command {
 
     pub async fn send(&mut self, message: &[u8]) -> Result<(), RadarError> {
         if self.sock.is_none() {
+            if !self.own_socket_allowed {
+                // Unicast topology, shared socket not yet available: drop the
+                // command rather than open a colliding own socket. The report
+                // loop retries socket creation and will supply the shared one.
+                return Ok(());
+            }
             self.start_socket().await?;
         }
         if let Some(sock) = &self.sock {
--- a/src/lib/brand/raymarine/command.rs
+++ b/src/lib/brand/raymarine/command.rs
@@ -1,4 +1,5 @@
 use async_trait::async_trait;
+use std::sync::Arc;
 use tokio::net::UdpSocket;
 
 use super::BaseModel;
@@ -15,7 +16,7 @@ pub(crate) struct Command {
     key: String,
     info: RadarInfo,
     model: BaseModel,
-    sock: Option<UdpSocket>,
+    sock: Option<Arc<UdpSocket>>,
 }
 
 impl Command {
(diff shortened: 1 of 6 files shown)


dirkwa:
@coderabbitai review

dirkwa:
@keesverruijt 

WOHOO!!!

The user reported it works!

image

The entire chat here:
https://discord.com/channels/1170433917761892493/1511825457115299961

As this is now deep changes I do not feel comfortable to merge this on my own. 
I would love you to review it, or even rewrite it as you would love to have it. 

We tested the above with 
`--allow-wifi --brand raymarine -v`

keesverruijt:
Ok good, I took a brief look at the code and it is good but not perfect, in particular some things can be made a little less complicated I think now (but maybe I'm wrong.) I will get to this next week when I am at sea again.

keesverruijt:
Rebased onto current `main` (conflict in `raymarine/mod.rs` test module — kept both test sets) and added six follow-ups while reviewing:

- **`fix(raymarine): collapse all RadarInfo addrs onto the radar in unicast mode`** — `info.addr` / `report_addr` / `spoke_data_addr` were `0.0.0.0:port`; they're all the radar's command address now. Multicast path unchanged so existing radar keys don't shift.
- **`refactor(raymarine): thread one unicast_mode flag through Command and the report receiver`** — replaces the two-step `new()` + `disable_own_socket()` and the two recomputations of `!report_addr.is_multicast()`. `Command.own_socket_allowed` becomes `Command.unicast_mode` (inverted polarity); asserts on `set_shared_socket` / `start_socket`.
- **`refactor(raymarine): start_report_socket can't fail, hoist back-off into run()`** — every exit was `Ok`; dropped the `io::Result<()>`, moved the back-off sleep into `run()`, single call site, early `continue`, inline drop.
- **`fix(raymarine): observe shutdown while backing off after a failed report-socket bind`** — `tokio::select!` the back-off against `subsys.on_shutdown_requested()` so a network outage can't hang graceful shutdown.
- **`style(raymarine)`** — `cargo fmt` on the `FeatureFlags` accessors and the model lookup table, plus `use` ordering.
- **`refactor: split RadarSocket into Udp(UdpSocket) and SharedUdp(Arc<UdpSocket>)`** — the `Arc` was only needed for the unicast topology; making the variant explicit removes the wrap from every other brand's listen path.

All 284 + 15 tests pass; `--features pcap-replay` clean.

keesverruijt:
## Rebased onto fresh main + CodeRabbit follow-ups

Branch rebased onto current `main`. Conflicts in `src/lib/brand/raymarine/{report.rs,report/quantum.rs}` resolved in favour of the PR's intent (unicast `SocketType::Any`, threaded `unicast_mode` flag, hoisted back-off in `run()`). Verified clean against:

- `cargo fmt --check`
- `cargo clippy --no-deps --all-targets -- -D warnings` (default + `pcap-replay`)
- `cargo test` — 305 lib + 15 integration tests pass

### CodeRabbit findings

| # | Finding | Status |
|---|---------|--------|
| 1 | `report.rs:310` log says "create multicast failed" but the socket is now `SocketType::Any` | **Fixed** — now logs "create UDP listen socket failed" |
| 2 | `mod.rs:380` collapses `RadarInfo.addr` to listen address when report is unspecified, causing key collisions | **Already fixed** by later commit `78b96e9` ("collapse all RadarInfo addrs onto the radar in unicast mode") — `radar_addr` resolves to the radar's actual command address in that branch |
| 3 | `create_connected_unicast()` failure falls back to split-socket | **Already fixed** by commit `0f9008c` ("retry unicast socket instead of falling back to a split socket") — the `Err` arm now logs and returns, leaving `report_socket` `None` so `run()` retries |

Both #2 and #3 are stale — CodeRabbit reviewed an early commit; the fixes landed later in the same PR.

PR #358, opened 2026-06-16
fix(clippy): make  warning-clean and gate it in CI + pre-commit

### LINKED ISSUE #355: Ensure all commits are clippy + fmt clean
Ensure no commits are pushed that aren't cargo clippy or cargo fmt clean.

keesverruijt:
Drives `cargo clippy --no-deps --all-targets -- -D warnings` from 344 warnings + 4 hard errors to **0**, then wires up CI and the pre-commit hook so it stays that way.

Both feature configs (default and `--features pcap-replay`) are clean. All 283 lib + 15 integration tests still pass.

## What landed

30 commits in three buckets:

- **Hard errors (4 fixes).** Never-looping `loop` in the emulator branch, `Ord/PartialOrd` mismatch on `Range`, and two `FRAC_PI_4` approximations in the OpenAPI doc examples.
- **Warnings (~320, in themed batches).** `collapsible_if`, `needless_return`, `clone_on_copy`, `needless_borrow`, `len_zero`, `useless_conversion`, `unnecessary_cast`, `redundant_closure`, `manual_clamp`, doc-list formatting, `box_collection`, `ptr_arg`, transmute annotations, every `needless_range_loop` rewritten to the appropriate iterator shape, `repr(packed)` → `repr(C, packed)` on every wire-format struct (26 sites; ABI explicitly locked), and per-site `#[allow]`s for the few categories where the refactor isn't a net improvement (`too_many_arguments`, `large_enum_variant`, `type_complexity`, `upper_case_acronyms`).
- **Gating.** New `clippy` job in `.github/workflows/rust.yml` runs both feature configs with `-D warnings`. `.githooks/pre-commit` mirrors it after the fmt step. `AGENTS.md` updated.

Closes #355.

## Quiet bug picked up along the way

`let_underscore_future` caught that the WS error-response path in `handle_client_request` was dropping `socket.send(...)` un-awaited — error replies to the client were never being sent. Now properly `.await`ed (commit `b56c9ed`).

## Test plan

- [x] `cargo clippy --no-deps --all-targets -- -D warnings` clean (default features)
- [x] `cargo clippy --no-deps --all-targets --features pcap-replay -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] `cargo build` and `cargo build --features pcap-replay` clean
- [x] `cargo test --no-fail-fast` — 283 lib + 15 integration, 0 failed
- [x] Pre-commit hook smoke-tested: returns 0 on clean tree, returns 1 with a focused error excerpt on dirty trees (both fmt and clippy paths exercised)

## Review note

CodeRabbit may flag two pre-existing soft regressions that didn't show up locally because they live behind `--features pcap-replay`:

- `args.clone()` was restored in `start_session` (commit `10393dc`) after the `clone_on_copy` auto-fix sweep dropped it. `Cli` isn't `Copy`, so the lint was a false positive; the clone is required by the pcap-replay branch that reads `args.repeat` after the move.

If you want the 30 commits flattened, `git rebase -i --autosquash origin/main` will collapse the per-category fix batches but I left them split because each is a separately reviewable diff.

Eliminates all clippy warnings and hard errors from the codebase by fixing approximately 320 warnings and 4 hard errors across both default and `--features pcap-replay` configurations, then gates the checks in CI and pre-commit hooks to maintain the clean state.

## Hard Error Fixes

Corrects a never-looping `loop` in emulator mode, resolves an `Ord`/`PartialOrd` contract violation in `Range` by implementing `Ord` manually (ordering solely by `distance` rather than deriving it), and removes two `FRAC_PI_4` approximations from OpenAPI documentation examples.

## Mechanical Warning Fixes

Applies clippy auto-fixes including `collapsible_if`, `needless_return`, `clone_on_copy`, `needless_borrow`, and `len_zero`. Converts `needless_range_loop` instances to appropriate iterator shapes, including refactoring loops to use `enumerate()`, `iter_mut()`, and zipped iterators across the IMM motion model, relative-trail rendering, and spoke processing logic.

Locks the ABI of 26 wire-format structs with `#[repr(C, packed)]` (updating `FurunoRadarReport`, `FurunoRadarModelReport`, `HaloHeadingPacket`, `HaloNavigationPacket`, and Raymarine/Navico packet structs), changing from `#[repr(packed)]` to ensure explicit C-compatible layout.

Applies targeted `#[allow]` annotations where refactoring provides no net improvement (e.g., `vec_init_then_push` with feature-dependent push elements, `upper_case_acronyms` for `RadarModel`, `large_enum_variant` for `Connection` enum).

## API and Semantic Changes

Changes `Range`'s derivation to implement `Eq` only with a manual `Ord` impl; refactors `Units` conversion methods (`to_si` and `from_si`) to take `self` by value instead of `&self`; replaces `impl Into<Brand> for &str` with `impl From<&str> for Brand`; switches `InterfaceApi` to use `#[derive(Default)]` instead of a manual impl; adds `impl Default for SharedRadars` and `impl Default for SignalKDelta`.

Changes ownership/borrowing at multiple call sites: methods like `CommonRadar::new` now take `radars` by value rather than cloning it, and several functions now pass `subsys`/handles by value instead of by reference throughout the brand-specific radar implementations.

Updates method signatures: `Control::set_valid_ranges` parameter type changes from `&Vec<Range>` to `&[Range]`; `TrailBuffer::relative_trails` field type changes from `Box<Vec<u16>>` to `Vec<u16>`.

## Refactoring Patterns

Removes unnecessary cloning throughout: address/socket assignments now dereference (`*addr`) instead of calling `.clone()`; error construction now uses `io::Error::other(...)` instead of `io::Error::new(io::ErrorKind::Other, ...)`; several numeric comparisons and bounds checks switch to idiomatic ranges (`.contains()`/`is_multiple_of()`/`.is_empty()` instead of manual conditions).

Simplifies control flow: nested `if let` blocks become combined `if let ... && let ...` patterns; explicit `match` statements convert to `matches!` where appropriate; loop indices switch to iterator-based forms with `enumerate()`.

## Bug Discovery

Identifies and fixes an un-awaited `socket.send()` call in the WebSocket error-response path in `src/bin/mayara-server/web/signalk/v2.rs` where error replies were silently dropped and never sent to clients, adding `.await` to ensure replies reach clients.

## CI and Developer Workflow Gating

Adds a new `clippy` job to `.github/workflows/rust.yml` running `cargo clippy` with `--no-deps --all-targets -D warnings` for both default and `pcap-replay` feature configurations.

Mirrors the clippy check in `.githooks/pre-commit` after the format check, blocking commits when clippy fails (matching existing `cargo fmt --check` behavior).

Updates `AGENTS.md` contributor instructions to require running both `cargo fmt` and `cargo clippy --no-deps --all-targets -- -D warnings` before commits that touch Rust code, documenting the new pre-commit hook behavior.

## Test Suite Status

283 lib + 15 integration tests pass; 0 failures. All integration tests updated to pass subsystem handles by value to `mayara::start_session`.


keesverruijt:
Pushed two follow-up commits:

- **`10f2796`** — CodeRabbit review:
  - **Critical** `trail.rs` `index_prev` reset per spoke (I introduced this in the `needless_range_loop` rewrite — the inverted-range `.fill()` would panic).
  - **Major** `rust.yml`: restrict the new `clippy` job's `GITHUB_TOKEN` to `contents: read`.
  - **Minor** `brand/mod.rs`: drop the echo comment I'd added.
- **`48f38da`** — close 8 clippy lints introduced by rustc 1.96 that didn't fire on my local 1.90 (`unnecessary_unwrap` ×5, `needless_range_loop` ×2, `redundant_closure`, `io_other_error`, `needless_return`, `unnecessary_sort_by`).

Both `cargo clippy --no-deps --all-targets -- -D warnings` and the `--features pcap-replay` variant pass on rustc 1.96 now. Full test suite still green.

### CodeRabbit findings I'm deferring to follow-up PRs (all pre-existing, out of this PR's scope)

- `radar/settings.rs::parse_path` mis-assigns the control segment to `radar_id` — multi-radar path decoding bug.
- `raymarine/report/quantum.rs::process_frame` and `process_spoke` need bounds checks on UDP packet length.
- `radar/settings.rs:880` `HashMap` iteration order used as enum value mapping.
- `stream.rs:691` throttle bug — `last_sent` advances even when the update is dropped.
- `emulator/report.rs:160` and `radar/mod.rs:896` magic numbers for `Power::{Standby, Transmit}`.

Happy to open issues / split into PRs for those — let me know if you want me to file them.

I'm also leaving the WS error-reply `.await` change inside this PR (CodeRabbit suggested splitting it). The fix is small, obviously correct (drop-without-await was a no-op bug), and the PR description already calls it out.

PR #367, opened 2026-06-16
fix(settings): changing controls via web socket did not work

### LINKED ISSUE #360: fix(settings): parse_path stores the control segment into radar_id
Flagged by CodeRabbit on PR #358.

**Location:** [`src/lib/radar/settings.rs:2051-2055`](https://github.com/MarineYachtRadar/mayara-server/blob/main/src/lib/radar/settings.rs#L2051-L2055)

\`\`\`rust
if let Some(r) = path.split('.').next_back() {
    self.control_id = ControlId::try_from(r).ok();
    if self.control_id.is_some() {
        self.radar_id = Some(r.to_string());   // ← wrong: r is the control name, not the radar id
        return Some(r);
    }
}
\`\`\`

\`r\` from \`path.split('.').next_back()\` is the **final** token (the control name, e.g. \`"gain"\`), not the radar id. \`self.radar_id\` ends up holding the control name, which breaks multi-radar path decoding.

### Suggested shape

\`\`\`rust
if let Some((radar_id, control_name)) = path.rsplit_once(".controls.") {
    self.control_id = ControlId::try_from(control_name).ok();
    if self.control_id.is_some() {
        self.radar_id = Some(radar_id.to_string());
        return Some(control_name);
    }
}
\`\`\`

### Verification

- A unit test that parses \`radars.nav1034A.controls.gain\` and asserts \`radar_id == "radars.nav1034A"\` (or whatever the prefix shape is) — none exists today as far as I can see.
- Manual: send a control PUT via the Signal K stream for a non-default radar and confirm it routes to the right one.

keesverruijt:
## Summary

`RadarControlValue::parse_path` was decoding the WS control path `radars.<radar>.controls.<control>` by taking the **last** dot-separated segment as the radar id. The last segment is the **control** name, so:

- `self.radar_id` was filled with the control name (\"gain\"), and
- the function returned the control name to the caller.

The WS stream handler then called `radars.get_by_key("gain")`, found nothing, and silently dropped the request. Every WebSocket control PUT was a no-op in multi-radar (and single-radar) setups. REST PUT was unaffected — it has its own routing.

The fix splits the path with `rsplit_once(".controls.")` so the radar id is the prefix and the control name is the suffix.

## End-to-end verification

Smoke-tested against a live mayara on a Pi with two Navico HALO24 heads (nav1034A + nav1034B):

| | OLD binary | NEW binary |
|---|---|---|
| WS `{path: radars.nav1034A.controls.gain, value: 77, auto: false}` | gain stayed at the previous REST-set value, same timestamp — silent drop | gain → **77 manual** immediately |
| nav1034B (untouched) | n/a | unchanged, auto — multi-radar isolation works |
| REST PUT (positive control) | works | works |

## Tests

Adds five unit tests for `parse_path`:

- canonical form `radars.<radar>.controls.<control>`
- some clients send the path without the leading `radars.` segment
- unknown control name → returns `None`, leaves fields unset
- missing `.controls.` separator → returns `None`
- hypothetical multi-segment radar id (e.g. `foo.bar`) — `rsplit_once` must pick the right-most separator

## Test plan

- [x] `cargo build` + `cargo build --features pcap-replay` clean
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --no-deps --all-targets -- -D warnings` and the `--features pcap-replay` variant clean
- [x] `cargo test` — 288 lib tests pass (was 283; +5 new)
- [x] Live WS smoke-test against the patched binary — old behaviour reproduced, new behaviour confirmed

Closes #360

## Bug Fix: WebSocket Radar Control Path Parsing

Fixes `RadarControlValue::parse_path` to correctly extract radar identifiers from WebSocket control paths. The previous implementation used `path.split('.').next_back()` to extract the radar ID from paths like `radars.<radar>.controls.<control>`, but this actually returned the final segment (the control name) instead of the radar identifier. This caused WebSocket control PUT requests to fail silently, as handlers attempted to look up radars by control names (e.g., `radars.get_by_key("gain")`) instead of radar IDs.

The fix replaces the parsing logic with `rsplit_once(".controls.")`, which correctly decomposes the path into the radar ID prefix and control name suffix. The implementation now also tolerates paths that omit the leading `radars.` prefix and preserves multi-segment radar IDs that contain dots.

The test suite is extended with five new unit tests covering canonical paths, paths without the `radars.` prefix, unknown control names, missing `.controls.` separators, and multi-segment radar IDs. Smoke testing on a live two-radar setup confirms WebSocket control PUTs now route to the correct radar while maintaining multi-radar isolation and REST functionality.

**Changes**: +97/-9 lines in `src/lib/radar/settings.rs` (internal logic and tests only; function signature unchanged).

### DIFF
--- a/src/lib/radar/settings.rs
+++ b/src/lib/radar/settings.rs
@@ -2043,17 +2043,27 @@ impl RadarControlValue {
         }
     }
 
+    /// Decode `path` into `radar_id` and `control_id`.
+    ///
+    /// The path is built by [`Self::new`] as `radars.<radar>.controls.<control>`.
+    /// On success the matching fields on `self` are populated and the borrowed
+    /// `radar_id` slice is returned so the caller can immediately look the
+    /// radar up.
     pub fn parse_path(&mut self) -> Option<&str> {
         let mut path = self.path.as_str();
-        if path.starts_with("radars.") {
-            path = &path["radars.".len()..];
+        if let Some(rest) = path.strip_prefix("radars.") {
+            path = rest;
         }
-        if let Some(r) = path.split('.').next_back() {
-            self.control_id = ControlId::try_from(r).ok();
-            if self.control_id.is_some() {
-                self.radar_id = Some(r.to_string());
-                return Some(r);
-            }
+        // Split on the canonical `.controls.` separator. Using `split('.')`
+        // and taking the last segment as the radar id was wrong: it stored
+        // the *control* name in `radar_id`, which broke multi-radar routing
+        // because the caller in the Signal K stream handler then looked the
+        // radar up by the control name and silently dropped the request.
+        let (radar_id, control_name) = path.rsplit_once(".controls.")?;
+        self.control_id = ControlId::try_from(control_name).ok();
+        if self.control_id.is_some() {
+            self.radar_id = Some(radar_id.to_string());
+            return Some(radar_id);
         }
 
         None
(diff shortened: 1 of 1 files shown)


keesverruijt:
Addressed in 1b76445 — `ControlId::try_from` -> `ControlId::from_str`, matching the existing style elsewhere in the file.

PR #429, opened 2026-07-04
fix(gui): stabilize north-up to true north via per-spoke bearing

dirkwa:
## Summary

North-Up mode did not hold true north. It rotated a bow-up image by a heading snapshot that was refreshed only on manual redraws (zoom, range, mode toggle), never when the boat's heading actually moved — so the image drifted away from north as the boat yawed at anchor, and an H↔N toggle snapped it back by the accumulated swing (tens of degrees). Separately, the AIS/ARPA overlay used a different heading source than the radar image, so targets drifted apart from their echoes.

- Write each spoke into the raster at its geographic `bearing` (already sent per spoke) instead of its relative angle plus a whole-image rotation. Every echo lands at its true direction the instant it arrives, so drift is impossible regardless of redraw cadence; the shader rotation is pinned to 0 and H↔N switches remap the existing rows in place to stay continuous rather than blanking for a sweep.
- Unify overlay math on live, draw-time heading with two explicit frames: the radar's own write-frame heading for anything glued to the echoes (guard/exclusion zones, no-transmit sectors, ARPA targets, MARPA acquire, compass rose, mouse hit-testing) and — for the true-bearing text readout — the true heading. AIS symbols are placed in the raster's write frame so they sit on their echoes even when the radar's heading feed disagrees with Signal K.
- North-up fills only the heading-induced part of a forward display gap, so a radar's natural angle stride (Furuno reports 8192 angles, sends every ~14th) is not mistaken for yaw and refilled on every spoke.
- A spoke with no usable heading is dropped rather than written bow-relative into the geographic raster; a sustained run auto-reverts to head-up. The radar-derived heading yields to a live Signal K heading once stale, and drops with flap tolerance so a marginal feed does not bounce the write frame mid-rotation.
- Fix `heading == 0` (due north) being treated as "no heading", re-upload the raster after renderer resize so the WebGL backend does not blank on a mode switch, and invalidate both Signal K headings on socket loss.

Reduce/Fill/Smooth processors take an optional display write-angle; rotation tracking and reduce calibration stay on the monotonic relative angle. Head-up behaviour is unchanged.

## Tested

- `cargo test` (329), `cargo fmt --check`, `cargo clippy -D warnings` all clean (change is JS-only).
- Headless browser (WebGPU and WebGL backends) against the emulator: bearing-indexed writes; geographic stability across a 90°→180° heading change; N↔H↔N raster remap continuity; heading-step gap fill; Furuno-stride gaps NOT filled; `heading == 0°` north-up allowed; heading-less spokes dropped and sustained loss auto-reverting to head-up; WebGL non-blank across mode toggles.
- AIS placement verified from the drawn pixel: with radar-derived 90° and Signal K true 60°, a true-north target renders straight up (screen angle 359.8°, on its echo) instead of the ~30° offset the split-frame formula produced.

North-up PPI rendering writes each spoke into the raster using its geographic bearing (via a per-spoke write-frame angle) instead of rotating the whole image from a stale heading snapshot, keeping north-up stable as heading changes and maintaining continuous H↔N toggles by remapping existing raster rows in place.

Radar overlays and interaction become heading-consistent at draw time: echo-linked elements follow the radar write-frame heading, true-bearing readouts and AIS placement use true heading, and target acquisition requires a live heading source. The renderer performs forward-display gap handling, drops headingless spokes, fills only targeted forward gaps (while avoiding natural stride gaps), auto-reverts to head-up after sustained heading loss, re-uploads the raster on renderer resize, and invalidates Signal K heading state on socket loss. Heading `0` is treated as valid north, and missing true heading displays as `HdgT: --°`.

Spoke processing adds a separate `writeAngle` from intrinsic `spoke.angle`, extending processor APIs to place buffer offsets and gap/fill calculations in display space for clean, fill, reduce, and smoothing processors while keeping calibration and related logic tied to the intrinsic geometry.

### DIFF
--- a/web/gui/ppi.js
+++ b/web/gui/ppi.js
@@ -12,13 +12,44 @@ const NAUTICAL_MILE = 1852.0;
 // Exponential-smoothing factor for the radar-derived heading. The bearing
 // field on each spoke is quantized, so the per-spoke `bearing - angle` heading
 // wobbles by a fraction of a degree frame to frame even with a steady boat.
-// In head-up mode the AIS/ARPA overlay subtracts this heading to place
-// geographic targets bow-up (the spoke raster is drawn bow-up and does NOT
-// subtract it), so the wobble shows up as targets jumping. Smoothing the
-// heading on its unit vector (circular-mean safe across the 360->0 wrap) kills
-// the jitter while staying responsive: at typical spoke rates the time
-// constant is well under the timescale of a real turn.
+// The overlay (AIS/ARPA/zones/rose) uses this smoothed heading to convert
+// between geographic and bow-relative angles, so the wobble would show up as
+// targets jumping. Smoothing the heading on its unit vector (circular-mean
+// safe across the 360->0 wrap) kills the jitter while staying responsive: at
+// typical spoke rates the time constant is well under the timescale of a real
+// turn. The north-up raster itself does NOT use the smoothed value — each
+// spoke is written at its own raw `spoke.bearing`, which is exact per spoke.
 const HEADING_SMOOTH_ALPHA = 0.1;
+
+// Overlay redraw threshold for live heading motion, in radians (~0.5°).
+// Below this the overlay keeps its last paint; above it a redraw is
+// scheduled so head-up AIS/ARPA placement and north-up zones/rose track
+// the boat's yaw even when no other repaint trigger fires (e.g. standby).
+const HEADING_REDRAW_EPSILON = (0.5 * Math.PI) / 180;
+
+// In north-up mode spokes are written at their geographic bearing, so the
+// display row can skip forward beyond the sweep's own step when the heading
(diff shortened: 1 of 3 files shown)


dirkwa:
@coderabbitai Good catch — fixed in c4279f0.

You're right that the radar-derived heading is a first-class source here: under per-spoke stabilization it's what places the north-up raster, so forcing head-up on Signal K socket loss dropped north-up even when the radar was still sending bearings. `onHeadingLost` now gates both the mode revert and the toggle-disable on `ppi.hasHeading()` being false — otherwise it keeps the current mode and just falls the PPI back to the radar heading (the invalidation of the Signal K readout is unchanged).

Verified against the real `onHeadingLost` path (closing the actual heading WebSocket in a headless browser): with the emulator's radar-derived heading live and no Signal K heading, north-up is preserved and the toggle stays enabled after the socket closes. The no-heading-remains branch still reverts to head-up as before.

PR #435, opened 2026-07-15
fix(target): bound blob detector memory for never-completing blobs

### LINKED ISSUE #434: Steady memory growth leads to OOM-kill roughly every ~19h (radar target tracker active, 0 tracked targets)
## Summary

`mayara-server` on our Raspberry Pi (arm64) HALPI2 boat computer has been getting OOM-killed by its container's 1GiB cgroup memory limit roughly every ~19 hours during continuous operation, with memory climbing steadily from a fresh-restart baseline the whole time. This looks like a genuine leak rather than a traffic-driven spike.

## Environment

- Image: `ghcr.io/marineyachtradar/mayara-server:main` (rolling tag, pulled 2026-07-05)
- Host: Raspberry Pi (HALPI2), arm64, Docker 26.1.5
- Invocation: `mayara-server -n ws:127.0.0.1:3000 --brand navico --interface eth0`
- Container: `network_mode: host`, memory limit 1GiB (cgroup), no explicit volumes/config file
- Connected radar: Navico (brand flag `navico`), actively scanning during a multi-day offshore passage

## Observed pattern

Two consecutive OOM kills, confirmed via `dmesg`:

```
[Wed Jul 15 01:33:29 2026] Memory cgroup out of memory: Killed process 2099 (mayara-server) total-vm:1217392kB, anon-rss:902624kB, ...
[Wed Jul 15 20:17:12 2026] Memory cgroup out of memory: Killed process 615508 (mayara-server) total-vm:1225440kB, anon-rss:988272kB, ...
```

That's **~18h 44m** apart. `docker stats` right after each auto-restart shows a fresh baseline around **165MiB**, so each cycle climbs roughly 165MiB → ~900-990MiB over ~19 hours before hitting the 1GiB limit — a steady, continuous climb, not a burst.

The container's own healthcheck/restart-policy recovers it automatically each time (`restart: unless-stopped`), so the practical impact is a brief radar-plugin restart roughly once a day rather than a hard outage — but it's clearly not supposed to be growing unbounded.

## What doesn't seem to explain it

- **AIS vessel counts are modest** — logged `Sending N AIS vessels` over the SignalK stream ranges from 1 to 53 over a 24h window, not obviously growing over time, and too small in absolute terms to plausibly account for hundreds of MB.
- **Target tracker shows 0 confirmed targets throughout** — the whole time leading up to and including the log window I could inspect, `mayara::radar::target::tracker` logs read like:
  ```
  Revolution 580: 0 targets (0 acquiring, 0 tracking), 34 candidates processed
  Revolution 581: 0 targets (0 acquiring, 0 tracking), 28 candidates processed
  Revolution 582: 0 targets (0 acquiring, 0 tracking), 63 candidates processed
  ...
  ```
  i.e. every revolution's candidate detections get evaluated and none are promoted to acquiring/tracking — so the confirmed-target store itself should stay essentially empty. That points away from "leaking retained target state" and toward something in the per-revolution candidate/spoke-buffer handling (raw blob candidates, spoke history, or similar) not being freed each cycle even when nothing is promoted.

## What I wasn't able to capture

- Logs from immediately before either OOM kill — the container isn't journald-logged on our setup, and `docker logs` only covers the current (post-restart) instance, so the previous instance's stdout is gone by the time I investigated. If there's a cheap way to get this container onto persistent logging (or if the maintainer wants a debug/trace build), I'm happy to run it and capture a full cycle.
- I did *not* have time to run this under `valgrind`/`heaptrack`/a Rust allocator profiler — happy to do so against our live radar if that would help narrow it down, given it's fully reproducible here (~19h cycle, no special trigger needed, just normal operation with a Navico radar transmitting).

## Checked first

- No existing open issue here mentions memory or OOM.
- Nothing in the last ~15 commits before our pulled build (`main` as of 2026-07-05) touches memory handling, so this doesn't look like something already fixed and just not yet in our image.

Happy to gather more data (longer memory-vs-time samples via `docker stats`, a profiler run, etc.) if that would help — this is fully reproducible on our end roughly once a day.

dirkwa:
On long passages with a scanning radar, mayara-server's memory grows steadily (~45 MiB/h with a Navico at ~24 RPM) until the process runs out of memory — in a 1 GiB container it gets OOM-killed roughly once a day, restarting the radar server mid-passage.

Root cause: the ARPA blob detector (on by default) only finalizes a blob after two spokes pass without extending it. A return present at every bearing — the near-range sea-clutter ring around own ship — never finalizes, and every revolution re-appended all of its pixels as duplicates, growing the blob's pixel list without bound (~13 KiB/s, matching the reported ~12.6 KiB/s).

The fix skips the duplicate push when the spatial index shows the pixel already belongs to the blob, refreshing only its liveness so completion semantics are unchanged. As a safety net, blobs exceeding `MAX_BLOB_PIXELS` are discarded, restoring the bound lost when the 1/4-revolution buffer cap was removed in the spatial-index rewrite (1ab52c7).

fixes #434

Tested: new unit tests reproduce the leak (the ring-blob test fails on current `main`), verify discrete targets still complete alongside a persistent ring, and cover the oversized-blob discard; `cargo fmt --check`, `cargo clippy --no-deps --all-targets -- -D warnings`, and `cargo test` (317 lib tests) all pass.

Prevents unbounded ARPA blob growth by refreshing existing `(spoke, pixel)` entries instead of adding duplicate pixels, and by discarding in-progress blobs that exceed `MAX_BLOB_PIXELS` while purging their associated entries from `pixel_index`.

Adds tests for persistent clutter rings, discrete targets, and oversized blobs to verify bounded memory use and correct blob completion behavior.

### DIFF
--- a/src/lib/radar/target/blob.rs
+++ b/src/lib/radar/target/blob.rs
@@ -535,19 +535,19 @@ impl BlobDetector {
                 .expect("target blob must exist");
             blob.has_doppler_approaching |= is_doppler_approaching;
             let key = (pixel.spoke, pixel.pixel);
-            let oversized = if self.pixel_index.get(&key).copied() == Some(target_id) {
+            if self.pixel_index.get(&key).copied() == Some(target_id) {
                 // The pixel already belongs to this blob from an earlier
                 // revolution — possible only for a blob that never completes,
                 // e.g. a clutter ring touching every bearing. Re-pushing it
                 // would grow `pixels` without bound; just refresh liveness.
                 blob.last_spoke_with_addition = spoke_angle;
-                false
             } else {
                 blob.add_pixel(pixel, spoke_angle);
                 self.pixel_index.insert(key, target_id);
-                blob.pixels.len() > MAX_BLOB_PIXELS
-            };
-            if oversized {
+            }
+            // Checked on both paths: a merge can push the blob over the cap
+            // even when the current pixel is a duplicate.
+            if blob.pixels.len() > MAX_BLOB_PIXELS {
                 let blob = self
                     .active_blobs
                     .remove(&target_id)
--- a/src/lib/radar/target/blob.rs
+++ b/src/lib/radar/target/blob.rs
@@ -24,6 +24,14 @@ const DEFAULT_BLOB_THRESHOLD: u8 = 10;
 /// bounding-box sizes; real vessels at this range produce dense clusters of 50+ pixels.
 const MIN_TARGET_PIXELS: usize = 25;
 
+/// Hard cap on the pixels retained by one in-progress blob. A blob that is
+/// extended on every spoke (e.g. a saturated clutter disk around own ship)
+/// never satisfies the completion check, so without a cap it would hold pixels
+/// for as long as the return persists. No plausible vessel return comes
+/// anywhere near this count (a valid target has at most a few thousand
+/// pixels); oversized blobs are discarded outright.
+const MAX_BLOB_PIXELS: usize = 100_000;
+
 /// Minimum ship size in meters
 pub const MIN_TARGET_SIZE_M: f64 = 5.0;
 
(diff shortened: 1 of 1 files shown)


dirkwa:
@coderabbitai review

PR #444, opened 2026-07-16
perf(target): reduce ARPA CPU usage ~3.5x

### LINKED ISSUE #442: refactor(target): replace pixel_index HashMap with a flat 2D array
## What users see

None directly — this is an internal refactor. Two indirect benefits:

- Lower per-lookup cost in the ARPA blob detector, which runs on every strong pixel of every spoke (the hot path when a radar is scanning).
- Fixes a latent bug where a mid-scan range change leaves stale spatial-index entries pointing at coordinates that no longer mean the same thing, so a subsequent adjacency check can incorrectly merge or split blobs after the operator changes range.

## Technical detail

Today `BlobDetector` keeps its spatial index in

\`\`\`rust
pixel_index: HashMap<(u16, usize), u32>,   // (spoke, pixel) -> blob id
\`\`\`

The key space is dense and bounded (\`spokes_per_revolution\` × \`current_spoke_len\`), so a HashMap is the wrong shape:

- Every insert / lookup hashes a \`(u16, usize)\` tuple.
- Each entry costs ~50 bytes of table overhead for a 4-byte payload.
- \`adjacent_blob_ids\` and \`calculate_contour\` do three or four neighbour hashings per pixel.
- The duplicate-pixel guard added in #435 (\`self.pixel_index.get(&key).copied() == Some(target_id)\`) is unnecessarily verbose vs. an array indexing.

Replace with a flat \`Vec<u32>\` sized \`spokes_per_revolution * current_spoke_len\`, indexed as \`row * current_spoke_len + col\`. Use \`0\` as an \"unowned\" sentinel and start \`next_blob_id\` at \`1\` (or use \`Option<NonZeroU32>\` for the same effect with type-level safety).

Memory footprint at 4 bytes/cell:

| Radar profile | Cells | Size |
|---|---|---|
| Navico 2048 × 1024 | 2 097 152 | 8 MiB |
| Furuno 4096 × 2048 | 8 388 608 | 32 MiB |

Fixed overhead regardless of scene complexity, but modest for the platforms mayara typically runs on and much smaller than the HashMap in a busy scene.

### Latent bug the refactor also fixes

\`BlobDetector::process_spoke\` currently updates \`current_range\` / \`current_spoke_len\` without resetting \`active_blobs\` or \`pixel_index\`. A range change mid-scan leaves stale \`(spoke, pixel)\` entries whose coordinates no longer refer to the same physical location; the next adjacency check can associate the wrong blob with a pixel. Reallocating the flat array whenever \`current_spoke_len\` changes forces the state to be cleared, so this can't survive.

### Scope

- Replace the field and the four call sites (\`adjacent_blob_ids\`, \`calculate_contour\`, the update paths in \`process_spoke\`, and the completion/discard cleanup).
- Reset the array and \`active_blobs\` on \`spoke_len\` change.
- Keep the per-blob \`Vec<BlobPixel>\` — that's still what \`SpokeArc::from_blob\` and cleanup iterate over.
- Extend the existing tests: no behavioural change is expected for the ring-blob or discrete-target cases; add a test that a mid-scan range change clears state.

Depends on #435 landing first (that PR fixes the OOM independently and shouldn't be gated on this refactor).

keesverruijt:
## What users see

When mayara has ARPA target detection enabled and radars are actively scanning, the server used a lot of CPU proportional to the number of returns being processed. On a Raspberry Pi 4 with two HALO24 radars scanning under `--targets arpa`, the process previously ran at ~66 % of one core; after this change it runs at ~19 %. On smaller Pis or busy configurations that were previously CPU-bound (or thermally throttling) the ARPA path now leaves plenty of headroom.

## Technical detail

Two per-spoke hot paths in `BlobDetector` dominated CPU. Both are fixed here because they hit the same loop and separating them would just double the reviewer overhead.

### 1. Flat spatial index

`pixel_index: HashMap<(u16, usize), u32>` → `pixel_index: Vec<u32>` sized `spokes_per_revolution × current_spoke_len`, indexed as `spoke * spoke_len + pixel`. `0` is a sentinel meaning "no blob owns this pixel"; blob ids start at 1.

The key space was dense and bounded, so a HashMap was the wrong shape: every insert / lookup hashed a `(u16, usize)` and every entry cost ~50 B of hashbrown overhead. The flat array costs ~8 MiB for a Navico (2048 × 1024 × 4 B) and is fixed regardless of scene complexity. Also fixes a latent bug: a mid-scan range change previously left stale `(spoke, pixel)` entries whose coordinates no longer meant the same physical location; the reallocated array forces the state to be cleared.

Closes #442.

### 2. Per-spoke scratch buffers

`process_spoke` previously allocated per call: a fresh `Vec<BlobPixel>` for strong pixels, a fresh `Vec<u32>` per adjacent-blob lookup, a fresh `Vec<u32>` for completed-blob ids, and `SpokeArc::from_blob` allocated + sorted a fresh `Vec<u16>` per completion. These are now detector-owned scratches taken via `std::mem::take` at the start of the call, mutated freely alongside `&mut self`, and put back at the end so their capacity survives. The strong-pixel find loop was fused with the processing loop, eliminating that `Vec` entirely.

## Measurements

Raspberry Pi 4 (Cortex-A72), two HALO24 radars, `--targets arpa`, 30 s / 99 Hz perf sampling with `--call-graph dwarf`:

|            | samples/s | CPU (of one core) |
|------------|-----------|-------------------|
| main       |        65 |               66% |
| step 1     |        27 |               27% |
| step 1 + 2 |      18.6 |               19% |

Allocator time (malloc + free + lock + unlock) dropped from ~24 % to ~3 % of total. HashMap symbols (`hash_one`, SipHash, hashbrown insert / remove) are gone from the profile entirely.

## Tests

- The 10 existing `BlobDetector` unit tests (blob completion under ring, discrete target alongside ring, oversized-blob discard, all `SpokeArc::from_blob` edge cases, guard-zone wrap) all pass.
- Added `spoke_len_change_resets_spatial_state` to lock in the latent-bug fix.
- No behavioural change is intended: the detector produces the same completed blobs from the same spoke inputs.

Replaces `BlobDetector`’s per-pixel `HashMap` spatial index with a dense flat `Vec<u32>` indexed by `spoke * spoke_len + pixel`, using `0` as the unowned sentinel and blob IDs starting at 1. Reuses per-spoke scratch buffers and fuses strong-pixel detection with blob processing to reduce CPU and allocator overhead. Resets detector state (active blobs and the flat index) when the spoke length changes to prevent stale ownership entries, while keeping behavior unchanged otherwise. Updates `SpokeArc::from_blob` and blob completion logic to use caller-provided scratch buffers, adjusts comments around duplicate-pixel handling and the fused loop, adds coverage for blob-ID `u32` rollover with an active ring, and updates tests to validate flat-index occupancy and spoke-length reset behavior.

### DIFF
--- a/src/lib/radar/target/blob.rs
+++ b/src/lib/radar/target/blob.rs
@@ -551,9 +551,8 @@ impl BlobDetector {
         let mut completed_ids = std::mem::take(&mut self.completed_ids_scratch);
         let mut spoke_arc_scratch = std::mem::take(&mut self.spoke_arc_scratch);
 
-        // Fused strong-pixel detection and processing: a single pass over
-        // spoke.data replaces the previous "build Vec<BlobPixel> then iterate"
-        // pattern, saving a per-spoke allocation.
+        // Strong/Doppler pixels are detected and processed in one pass over
+        // spoke.data so the hot path needs no intermediate pixel buffer.
         let doppler_range = self.doppler_approaching_range;
         for (pixel_idx, &intensity) in spoke.data.iter().enumerate() {
             let is_doppler_approaching = doppler_range
@@ -609,10 +608,11 @@ impl BlobDetector {
             let (pxl_spoke, pxl_pixel) = (pixel.spoke, pixel.pixel);
             let cell = pxl_spoke as usize * self.current_spoke_len + pxl_pixel;
             if self.pixel_index[cell] == target_id {
-                // The pixel already belongs to this blob from an earlier
-                // revolution — possible only for a blob that never completes,
-                // e.g. a clutter ring touching every bearing. Re-pushing it
-                // would grow `pixels` without bound; just refresh liveness.
+                // The pixel is already recorded in this blob: a long-lived
+                // blob re-touched on a later revolution (e.g. a clutter ring
+                // touching every bearing), or a merge above just reassigned
+                // this cell to the survivor. Re-pushing it would grow
+                // `pixels` without bound; just refresh liveness.
                 blob.last_spoke_with_addition = spoke_angle;
             } else {
                 blob.add_pixel(pixel, spoke_angle);
--- a/src/lib/radar/target/blob.rs
+++ b/src/lib/radar/target/blob.rs
@@ -474,6 +474,22 @@ impl BlobDetector {
         }
     }
 
+    /// Allocate a blob id. Wraps on u32 exhaustion, skipping `UNOWNED` and
+    /// ids still present in `active_blobs`: a wrapped id that aliased a live
+    /// blob would overwrite its `active_blobs` entry and strand that blob's
+    /// `pixel_index` cells on a dead id. Long-lived blobs (a clutter ring
+    /// touching every bearing) can hold a low id indefinitely, so low ids
+    /// are not safe to reuse blindly.
+    fn alloc_blob_id(&mut self) -> u32 {
+        loop {
+            let id = self.next_blob_id;
+            self.next_blob_id = self.next_blob_id.wrapping_add(1);
+            if id != UNOWNED && !self.active_blobs.contains_key(&id) {
+                return id;
+            }
+        }
+    }
+
     /// Flat-array index for `(spoke, pixel)`. Callers guarantee both fit
     /// within `spokes_per_revolution` and `current_spoke_len`.
     fn idx(&self, spoke: u16, pixel: usize) -> usize {
@@ -556,8 +572,7 @@ impl BlobDetector {
 
             let target_id = match adjacent_ids.len() {
                 0 => {
-                    let id = self.next_blob_id;
-                    self.next_blob_id += 1;
+                    let id = self.alloc_blob_id();
                     let mut blob = BlobInProgress::new(id, pixel.clone());
                     blob.has_doppler_approaching = is_doppler_approaching;
                     self.active_blobs.insert(id, blob);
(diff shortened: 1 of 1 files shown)


dirkwa:
Nice speedup — the flat index is clearly the right shape for this key space, and it keeps the #435 duplicate-pixel guard and its tests intact. I pushed two commits on top rather than leaving review comments, since both were quicker to show than to describe:

**7bab6ba — guard blob id allocation against u32 rollover.** `next_blob_id += 1` overflows after 2^32 allocations: debug builds panic, and release builds wrap to 0 — which now aliases `UNOWNED` — and then to low ids that can collide with a still-active blob. A collision overwrites that blob's `active_blobs` entry and strands its `pixel_index` cells on a dead id, so the next adjacency hit panics on `expect("target blob must exist")`. It's not just theoretical: an immortal clutter-ring blob holds a low id indefinitely, and heavy rain clutter can allocate on the order of a thousand ids per revolution, which brings 2^32 within ~50 days of continuous scanning. The allocator now wraps and skips `UNOWNED` and live ids — one u32 map probe per blob creation, nothing on the per-pixel path. The regression test parks the counter at `u32::MAX` with a ring alive and scans across the wrap; it panics on the previous code via the debug overflow check.

**1411e1e — comment accuracy.** CodeRabbit is right on both counts: the duplicate-pixel branch is also reached when a same-spoke merge reassigns the cell to the survivor (not only across revolutions), and the fused-loop comment described the old implementation rather than the current behavior.

One thing I deliberately did not touch: a range change that keeps `spoke_len` constant (the Navico case) still doesn't reset in-progress blobs. Cell space stays self-consistent there, so the only artifact is a one-revolution size blur on blobs alive across the change — clearing `active_blobs` on `range_changed` would be a two-line strictness win if you want it, but I didn't want to widen the PR.

Full chain is green locally on 1.97: fmt, clippy (both feature configs), 332 lib tests.

Issue #447, opened 2026-07-18
Second Furuno DRS radar never appears (NavNet 3D era units, empty serial number)


NaatuMik:
With two Furuno DRS radars on board (DRS6A and DRS12A, original NavNet 3D
magnetron generation), mayara only ever shows one radar. Both radars work
simultaneously in TimeZero and on the NavNet MFDBB. On Windows, sometimes no
radar is found at all even with one unit powered.

Findings from live testing and packet capture on the boat network (172.31/16):

1. Both units send the 170 byte model report with an all zero serial number
   field. `c_string()` turns that into an empty (not absent) string, so both
   radars get the bare key `fur` and `SharedRadars::add()` silently drops the
   second one. Fix in PR #448.

2. On Windows, any UDP recv error permanently ends discovery for a brand:
   the receive task is not respawned and the socket is dropped. Two errors
   actually occur in the field. WSAEMSGSIZE: UDP 10010 is shared by the whole
   NavNet ecosystem and we observed 378 byte nav data broadcasts, larger than
   the 300 byte locator buffer (Unix truncates, Windows errors). WSAECONNRESET:
   ICMP port unreachable echoed back from the beacon request broadcast.
   Discovery working at all was a race between the radar reply and the first
   such error. Fix in PR #449.

3. Discovery stops after the first radar unless `--multiple-radar` is given.
   The second unit's model report can arrive minutes after the first one, so
   the default behavior guarantees the second radar is missed. Is there a
   reason not to keep locators running by default (or at least for a grace
   period)? No PR yet, happy to discuss.

Identity observations for these units, relevant for keying and naming:

- The model field contains `DRS_OPEN` for both units. This is the NavNet
  nickname, not a model; it merely passes the `starts_with("DRS")` check.
- The serial number field is all zeros on both units.
- The units DO differ at bytes 19..23 of the model report:
  `1d 05 79 03` vs `1d 05 70 45`. Currently unparsed filler; candidate
  stable unit id.
- The 30 byte beacon report carries distinct names `RD000003` / `RD000004`.
  These are only debug logged today; discovery never correlates them with
  the model report (`half_found` is never populated on this path).
- Radar IPs are not stable: one unit moved 172.31.254.30 -> .32 -> .28
  across three days, so the IP fallback key changes and per radar settings
  do not follow the radar. IP keying fixes the collision but a stable
  identity (beacon name or bytes 19..23) would be better long term.
- `$N96` Modules response, identical on both units (software program
[...]

PR #448, opened 2026-07-18
fix(radar): key by IP when serial number is empty

NaatuMik:
With two Furuno DRS radars of the NavNet 3D generation on board, only one
radar appears in mayara. These units report an all zero serial number, so
both radars derive the same key `fur` and the second one is silently dropped
by `SharedRadars::add()`.

Treat an empty but present serial number the same as an absent one, so the
key falls back to the IP derived suffix and each radar gets a distinct key.

Fixes the collision part of #447.

Updates radar key generation to ignore empty serial numbers, allowing Furuno DRS radars to use IP-derived suffixes and avoid duplicate keys when multiple NavNet 3D radars are present.

### DIFF
--- a/src/lib/radar/mod.rs
+++ b/src/lib/radar/mod.rs
@@ -363,6 +363,29 @@ pub struct RadarInfo {
     is_idle: Arc<AtomicBool>,
 }
 
+/// Build the stable per-radar key: the brand prefix, a four-character
+/// discriminator, and the optional dual-range suffix. The discriminator is the
+/// last four characters of the serial number, or the low 16 bits of the radar
+/// IP when the serial number is absent *or* empty — so radars reporting an
+/// all-zero serial (e.g. Furuno NavNet 3D DRS units) fall back to the IP and
+/// each gets a distinct key instead of colliding on the bare prefix.
+fn radar_key(
+    prefix: &str,
+    serial_no: Option<&str>,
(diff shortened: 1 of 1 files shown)


keesverruijt:
Hi @NaatuMik, thanks for your first (and 2nd and 3rd) contributions! I will be addressing the CodeRabbit reviews and merge these ASAP.

keesverruijt:
Addressed the review in 23e85bb.

**CodeRabbit finding (add tests):** Extracted the key derivation into a pure `radar_key()` helper and added boundary regression tests:
- non-empty serial → last four characters (`fur2452`), and short serials used in full;
- `Some("")` and `None` → IP-derived suffix (`fur0102`);
- dual-range suffix appended, so two empty-serial radars on the same IP stay distinct (`fur0102A` / `fur0102B`).

**My own review — one additional fix:** the original change filtered the empty serial only when building the key, but line 415 still stored `serial_no: Some("")` in the struct. Several brands guard on `radar_info.serial_no.is_some()` and then use the value (e.g. navico/raymarine `settings.rs` do `serial.split_off(7)`), so a present-but-empty serial is a latent hazard. I now normalize the empty serial to `None` once, so the key **and** the stored field agree.

Build, tests, clippy (`-D warnings`) and fmt all clean.

PR #449, opened 2026-07-18
fix(locator): survive UDP recv errors on Windows

NaatuMik:
On Windows, radars are sometimes never discovered even though they are on
the network and beaconing: whether discovery works is a race between the
radar's reply and the first UDP recv error on the locator socket, because a
failed receive task is never respawned and its socket is dropped.

Two recoverable Windows specific errors occur in the field on a real boat
network. WSAEMSGSIZE: Furuno discovery listens on UDP 10010, which the whole
NavNet ecosystem broadcasts on, and observed 378 byte nav data broadcasts
exceed the 300 byte buffer (Unix truncates, Windows fails the recv).
WSAECONNRESET: ICMP port unreachable echoed back from a beacon request
broadcast fails the next recv.

Grow the buffer to 2048 bytes and keep receiving on these two recoverable
errors instead of dropping the socket.

Fixes the Windows discovery part of #447.

## Summary

- Increases the locator UDP receive buffer from 300 to 2048 bytes to accommodate larger broadcasts and avoid truncation.
- Continues receiving after recoverable Windows UDP errors (`WSAEMSGSIZE` and `WSAECONNRESET`) instead of dropping the socket.
- Handles IPv4 and IPv6 sources explicitly and reports other receive failures as I/O errors.

### DIFF
--- a/src/lib/locator.rs
+++ b/src/lib/locator.rs
@@ -28,6 +28,9 @@ use crate::{Brand, Cli, InterfaceApi, InterfaceId, InterfaceStatus, RadarInterfa
 // instead of being truncated.
 const LOCATOR_PACKET_BUFFER_LEN: usize = 2048;
 
+// Windows raw OS error for a datagram larger than the recv buffer.
+const WSAEMSGSIZE: i32 = 10040;
+
 // An Axiom wakes a radar with a burst of WOL packets roughly 20 ms apart, not
 // a single datagram — multicast to a dozing WiFi radar is lossy, so singles
 // are routinely missed. Space consecutive beacon-request packets the same way.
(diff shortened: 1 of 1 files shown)


keesverruijt:
@coderabbitai review

keesverruijt:
Addressed both review findings in 9925c9e:

- **IPv6 packets no longer kill discovery**: an IPv6 source packet is now logged at debug and skipped, keeping the receive socket alive, instead of returning `InterfaceNoV4` (which terminated the listener task and dropped the socket). `InterfaceNoV4` remains for local interface enumeration only.
- **Magic number `10040` → named `WSAEMSGSIZE` constant**.

PR #450, opened 2026-07-18
chore(build): drop unused vendored openssl dependency

NaatuMik:
Building from source on Windows fails in the vendored openssl build unless
Perl is installed. The openssl crate is not referenced anywhere in the
source tree or build script, and reqwest already uses rustls. Removing it
makes a plain `cargo build` work on Windows with only cargo and cmake.

Removes the unused vendored OpenSSL dependency and disables default features for development `reqwest`, preventing `native-tls` and `openssl-sys` from being pulled into builds and tests. Retains rustls for build-time requests, enabling Windows source builds without Perl.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -774,15 +774,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "48c757948c5ede0e46177b7add2e67155f70e33c07fea8284df6576da70b3719"
 
 [[package]]
-name = "encoding_rs"
-version = "0.8.35"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "75030f3c4f45dafd7586dd6780965a8c7e8e285a5ecb86713e63a79c5b2766f3"
-dependencies = [
- "cfg-if",
-]
-
-[[package]]
 name = "enum-primitive-derive"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


keesverruijt:
Confirmed this is safe to drop, with historical context for anyone wondering what the vendored openssl was doing there.

## Why it existed
The vendored openssl was added in commit `281fb6df` ("Fix build on musl linux", 2024-09-17), in the same commit that switched `reqwest` to `default-features = false` + `rustls-tls`. It was never used by our code — `vendored` forced `openssl-sys` (transitively present at the time) to compile from source so static **musl** linking wouldn't fail on the missing system libssl.

## Why it's safe now
`[dependencies]` and `[build-dependencies]` reqwest are both rustls-only, so the release binaries (musl / windows / macOS, all `cargo build --release`) never pull openssl. On Windows `native-tls` resolves to schannel and on macOS to Security.framework, so the explicit vendored build-dep was the *only* thing requiring Perl there — exactly the pain this PR removes.

## One extra thing I noticed
`openssl-sys` didn't vanish entirely: on **Linux** it was still pulled by the **dev-dependency** reqwest (default features → `hyper-tls` → `native-tls` → `openssl-sys`), which the vendored build-dep had been masking via feature unification. After this PR, `cargo test` / `cargo clippy --all-targets` on Linux would link *system* openssl instead. CI stays green (ubuntu has libssl-dev), but I pushed a follow-up commit tightening that dev-dep to `default-features = false` — the integration tests only hit `http://localhost`, so no TLS backend is needed. `cargo tree -i openssl-sys` / `-i native-tls` now return nothing on every platform, and the full test suite passes.

PR #455, opened 2026-07-19
feat(radar): show powered-off radars as off when they go silent

### LINKED ISSUE #432: GUI never shows a powered-off (red) radar state
## What a user sees

When a Raymarine radar is powered fully off, the GUI power icon does not show it as off (red) — it stays on its last colour (amber/standby or green/transmit), even after a page refresh. Amber and green track Standby and Transmit correctly; only the powered-off/red state is never shown.

Reported on a live boat in #160.

## Root cause

The power icon reflects the last power value the radar reported. A powered-off radar stops sending status reports, so mayara keeps the last value it saw and the icon never updates to off. mayara has no notion of "the radar has gone silent, treat its state as off/unknown."

This is the radar status/staleness model: when reports stop arriving for a radar, its power state should decay to a distinct "off/unknown" state that the GUI can render (red), rather than holding the stale last value. Related design discussion lives in the AGENTS.md "Radar status model" section.

## Scope

Separate from #427 (which fixed discovery, wake, control and power-off for Raymarine-behind-an-Axiom). This is a cross-brand status-model change, not Raymarine-specific.

keesverruijt:
When a radar is powered fully off, the GUI power icon used to keep showing its last state (amber/standby or green/transmit) forever — even across a page refresh — because a powered-off radar stops sending reports and mayara held the last value it saw. Now a radar that has gone silent is shown as off (red), and turns amber/green again as soon as it comes back.

Verified on **Navico**. **Garmin still needs on-water testing** — its report cadence in Standby is unconfirmed, so a genuinely silent Garmin standby could read as off.

Closes #432.

## Approach

This is the cross-brand status/staleness model from #432, not a per-brand fix:

- Each radar carries a shared last-input timestamp, bumped whenever its **report socket** receives a packet. Reports are the reliable heartbeat — a standby radar still sends periodic status, a powered-off one sends nothing. Hooked into Navico, Furuno (both dual-range partners), Raymarine, and Garmin. Koden is intentionally excluded: its broadcast recv socket isn't a trustworthy per-radar liveness signal.
- A central **watchdog** subsystem runs every 5s and decays any radar silent for ≥30s to `Power::Off`, pushing the change to the GUI/Signal K exactly like a real report. The decay applies from any live state, including Standby.

The 30s threshold assumes a standby radar reports at least that often (true for Navico/Raymarine/Furuno).

## Test plan

- [x] Unit tests for the `should_power_off` decision (before timeout, Standby→Off and Transmit→Off after timeout, already-off no-op, never-reported→Off)
- [x] Verified on Navico hardware (powered-off radar now shows red)

**Follow-up before release:** confirm Garmin standby report cadence on real hardware so a silent standby Garmin isn't misreported as off.

### DIFF
--- a/src/lib/brand/furuno/report.rs
+++ b/src/lib/brand/furuno/report.rs
@@ -310,6 +310,12 @@ impl FurunoReportReceiver {
                     match r {
                         Ok(len) => {
                             if len > 2 {
+                                // The report stream is shared by both ranges of a
+                                // dual-range radar, so a report keeps both alive.
+                                self.common.info.mark_input();
+                                if let Some(ref cb) = self.common_b {
+                                    cb.info.mark_input();
+                                }
                                 if let Err(e) = self.process_report(&line) {
                                     log::error!("{}: {}", self.common.key, e);
                                 } else if !first_report_received {
--- a/src/lib/brand/garmin/report.rs
+++ b/src/lib/brand/garmin/report.rs
@@ -333,6 +333,7 @@ impl GarminReportReceiver {
                 } => {
                     match r {
                         Ok((_len, _addr)) => {
+                            self.common.info.mark_input();
                             if let Err(e) = self.process_report(&report_buf) {
                                 log::error!("{}: {}", self.common.key, e);
                             }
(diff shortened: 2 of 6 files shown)


keesverruijt:
@coderabbit review

PR #457, opened 2026-07-20
fix(trail): correct target-trail colours

keesverruijt:
Enabling target trails painted them with the wrong colours: the empty background filled with white/grey, and freshly-laid trails showed up purple/green (the Doppler colours) instead of fading white→grey. This fixes all of it.

## Root causes

Three separate defects, all in the trail buffer:

1. **Whole background painted.** The "is this an echo worth trailing" threshold (`minimal_legend_value`) defaulted to `0`, because the `DopplerTrailsOnly` list control has no value yet when the buffer is built. That made the test `data >= 0` — always true — so every transparent no-echo pixel got a trail seeded and painted with a history colour. It's now initialised from the legend (the strong-return threshold) and can never be 0.

2. **Fresh trails rendered purple/green.** The buffer cached the legend at construction. On a Navico HALO, Doppler is detected *after* the radar is created, which rebuilds the legend and pushes `history_start` up past the newly inserted Doppler entries. The stale cached copy kept the old (lower) `history_start`, so fresh trails (low index) were written straight onto the Doppler colours. The cached legend is now refreshed whenever `history_start` changes.

3. **True-motion index overflow.** The true-motion path computed `trail * BLOB_HISTORY_COLORS` in `u8`, which overflows once the trail age reaches 8 (and panics in debug). Computed in `u16` now, matching the relative-motion path.

## Notes

Verified on a live Navico HALO: trails now lay down white and fade through grey over the selected trail duration, with a black background. Regression tests cover the threshold (never 0, correct strong/Doppler values).

## Summary

- Initializes trail echo thresholds from the legend to prevent empty backgrounds from being painted.
- Refreshes the cached legend when `history_start` changes, ensuring history trails use the correct white-to-grey colours.
- Uses `u16` intermediates for true-motion history indices to prevent overflow on older trails.
- Ensures normal-mode thresholds never return zero and adds regression coverage for threshold behaviour.
- Verifies trail fading on a live Navico HALO.

### DIFF
--- a/src/lib/radar/trail.rs
+++ b/src/lib/radar/trail.rs
@@ -51,13 +51,18 @@ impl TrailBuffer {
         );
 
         let legend = info.get_legend();
-        let mut minimal_legend_value = 0;
-        if let Some(control) = info.controls.get(&ControlId::DopplerTrailsOnly)
-            && let Some(value) = control.value
-        {
-            let value = value > 0.;
-            minimal_legend_value = Self::compute_minimal_legend_value(&legend, value);
-        }
+        // Lowest legend index that counts as an echo worth trailing: the
+        // strong-return threshold normally, or the first Doppler index when
+        // DopplerTrailsOnly is on. This must never be 0 — the transparent
+        // no-echo value (0) would then seed a trail on every empty pixel and
+        // paint the whole background with history colours.
+        let doppler_only = info
+            .controls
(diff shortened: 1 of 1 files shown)


keesverruijt:
@coderabbitai review

keesverruijt:
@coderabbitai full review

PR #469, opened 2026-07-22
fix(docker): add cmake required by libz-sys zlib-ng build 

night199uk:
When building the Docker image, cmake seems to be missing from rust:1.94-alpine arm64 image. 
Install it.

Adds CMake to the Docker builder image’s Alpine packages, ensuring Rust ARM64 builds have the required tool available.


dirkwa:
Thanks, could you please address the CR review, so I can merge it?

night199uk:
I have no idea what the CR review is asking for... cmake is missing from the alpine container which means the build is a complete failure, and adding cmake to the alpine container fixes the issue...

PR #478, opened 2026-07-27
feat(output): Garmin xHD output bridge

matztam:
## Summary

- Adds a new `garmin-xhd-output` Cargo feature that emulates a Garmin GMR xHD radar on the local Garmin Marine Network
- A Garmin GPSMAP chartplotter discovers the virtual radar automatically and displays spoke data from **any** radar source that Mayara supports (Furuno, Navico, Koden, emulator, PCAP replay)
- Controls flow in both directions: adjusting gain, sea clutter, rain clutter, range, or transmit state on the Garmin plotter sends the corresponding command to the real radar via Mayara's normal control system

## Architecture

Four concurrent tasks emulate the xHD radar:

| Task | Direction | Address | Port | Description |
|------|-----------|---------|------|-------------|
| CDM heartbeat | out | `239.254.2.2` | 50050 | Radar discovery (CDM V2 `0x038e`) |
| Status stream | out | `239.254.2.0` | 50100 | Settings/state broadcast (~1 Hz) |
| Spoke sender | out | `239.254.2.0` | 50102 | Sweep data (`0x0998` spoke packets) |
| Command listener | in | `<nic_ip>` | 50101 | Control commands from the plotter |

All `MSG_*` / `STATE_*` / address constants are reused from `src/lib/brand/garmin/protocol.rs` — none duplicated.

The feature is gated: `cargo build --release --features garmin-xhd-output`. Currently `spawn()` is called from the Furuno brand's `found()` method; other brands need their own call site.

## Key implementation notes

- **IP detection**: detects the Garmin NIC IP by filtering `172.16.0.0/12` addresses on the same NIC as the Furuno radar, avoiding Docker bridge addresses (`172.17–20.x.x`)
- **Socket source port binding**: each sender socket binds to its own well-known port (50050/50100/50102), not an ephemeral port — the plotter uses the source port for radar identification
- **Two range fields in spoke**: `range_meters` = nearest xHD table value (UI display), `display_meters` = raw `spoke.range` (chart overlay scaling). These must differ for correct overlay scaling.
- **Furuno range ratio**: Furuno's internal `spoke.range` ≈ 1.7822× the xHD range; corrected before snapping to table, raw value passed as `display_meters`
- **Rain state caching**: Mayara maps "rain off" to `auto=true` while leaving `value` unchanged; rain mode/gain are cached in `SharedState` so the plotter sees `rain_mode=0, rain_gain=0` immediately
- **Transmit state**: read directly from `SharedControls::Power` (updated by Furuno report parsing), with `SharedState.transmitting=false` as startup fallback
- **Range lock**: 5 s after plotter sets range, spoke thread ignores incoming `spoke.range` updates to prevent the old range from overwriting the new setting

## Documentation

- `docs/garmin-xhd-output.md` — end-user setup guide (network config, protocol overview, troubleshooting)
- `docs/internals/garmin-xhd-output.md` — developer internals (architecture, shared state design, spoke conversion, command routing)

## Test plan

- [ ] `cargo build --features garmin-xhd-output` compiles without errors
- [ ] `cargo test --features garmin-xhd-output` — all unit tests pass (angle mapping, range table, resampling)
- [ ] Garmin GPSMAP chartplotter discovers the virtual radar on the Garmin Marine Network
- [ ] Radar image visible and chart overlay correctly scaled at all ranges
- [ ] Range +/- buttons on plotter update the radar range
- [ ] Transmit on/off toggle works in both directions
- [ ] Gain, sea clutter, rain clutter controls work
- [ ] Rain clutter "off" correctly resets to 0% on plotter
- [ ] Standby/transmit state correctly reflects radar state at startup

Adds a gated `garmin-xhd-output` Cargo feature that emulates a Garmin GMR xHD radar over the local Garmin Marine Network.

- Defines `garmin-xhd-output` as enabling the existing `furuno` integration, and conditionally spawns the xHD output bridge from Furuno radar discovery.
- Selects an appropriate local IPv4 address on the configured radar NIC for multicast binding, then uses well-known UDP endpoints for CDM discovery heartbeats, periodic xHD status/range-table multicast, continuous xHD spoke multicast, and unicast command listening with Garmin acknowledgements.
- Implements four coordinated Tokio tasks with graceful shutdown: CDM heartbeat broadcaster, status/range-table broadcaster, paced spoke sender, and a plotter command listener that translates incoming Garmin TLVs into Mayara control updates (and ignores unsupported modes like RPM).
- Converts Mayara/Furuno spoke data into Garmin xHD `0x0998` spoke packets by snapping to the supported xHD range table, quantizing/wrapping angles, resampling intensity samples to a fixed count, applying Furuno-specific range correction, and using separate `range_meters` vs `display_meters` semantics.
- Maintains shared runtime state for current range (with a temporary range-lock window after range changes), transmit/power state, and cached rain mode/gain so broadcasts and command translations remain consistent.
- Buffers incoming spoke batches in a `VecDeque` and emits spokes at a steady microsecond cadence (dropping oldest packets when the queue grows) to prevent display coverage gaps when spokes arrive in bursty bursts.
- Adds end-user setup/troubleshooting docs for `garmin-xhd-output`, and developer internals documentation describing the bridge architecture, module layout, protocol behavior, supported command IDs, and shutdown/lifecycle ordering.
- Adds unit tests for xHD range snapping and spoke conversion/resampling, and introduces a Windows build workflow that builds with `garmin-xhd-output` enabled.


matztam:
## Spoke buffering — fix for missing sectors

During testing on hardware (Furuno DRS Radome + Garmin GPSMAP chartplotter) we discovered that the plotter was showing missing sectors (gaps in the radar sweep). After extensive debugging we identified the root cause:

**The problem:** Furuno delivers spokes in batches of ~1500 every ~450ms. Without pacing, our code sent all 1500 spokes as fast as possible (a few milliseconds), then went silent for ~450ms while waiting for the next batch. During that silence the plotter continued sweeping and rendered the gap as empty sectors.

**The fix:** We now pre-convert each incoming batch into a `VecDeque` and drain it at a steady rate of one spoke per 200µs. This matches roughly 2/3 of the real xHD spoke rate (~305µs at 24 RPM / 8192 spokes/rev), which means:
- The queue drains faster than Furuno fills it, so display lag stays low (~150ms average, ~22° angular offset)
- The next batch arrives before the queue runs empty, so the sweep is continuous
- At 150µs the queue occasionally ran dry and sectors reappeared — 200µs is the stable sweet spot

This approach is analogous to a jitter buffer in VoIP: absorb the bursty input, emit at a steady rate.

matztam:
Regarding the suggestion to extract the range-lock logic into a pure helper with unit tests: the current inline code in `handle_batch` is already concise (6 lines, one branch) and the three cases (lock active, lock expired + changed, lock expired + unchanged) are straightforward enough that tests would add more boilerplate than confidence. Keeping it inline avoids the indirection without sacrificing readability. Declining this suggestion.

dirkwa:
@coderabbitai resume

dirkwa:
@keesverruijt 

As we shortly spoke already about, this is a rater large PR, so I would love to have a 2nd opinion from you.

keesverruijt:
I am going to deny this PR, and would like your feedback on the (improved) implementation in #493.

Issue #54, opened 2026-04-07
Dual range radars: Side-by-Side PPI


dirkwa:
Some modern radars support dual range, for this we would beed a side-by-side PPI.

A research has to be done about the limitations per brand and model. 

F.e. the Furuno DRS4D-NXT is limited to 12nm on radar A and 6nm on radar B, also the controls (Gain, Bird etc) apply to both.

Once this is clear we would need to find a useful capabilities abstraction pattern that can be used in the API and UI.

keesverruijt:
This may be simpler than I thought: we can just open a web page that uses two frames, and then opens the normal GUI twice, each in that frame.

Colug0:
I have a dual range B&G 4G Radar, happy to supply feedback if needed.

PR #479, opened 2026-07-27
feat(gui): side-by-side view for dual-range radars

dirkwa:
Dual-range radars (Furuno NXT, Navico 4G/HALO) appear in mayara as two radars sharing one antenna, but until now they could only be watched one at a time — and on a Furuno NXT the second range never showed a picture at all. This adds a side-by-side dual view and fixes everything that kept range B dark, verified live against a DRS4D-NXT with both ranges painting and range B commanded from mayara (range confirmed by the radar's report loop).

- The radar list pairs the ranges of one antenna (new optional `dual` and `dualGroup` fields, documented as mayara extensions in the API-parity allowlist) and links a new `dual.html` that shows both ranges as two full viewers side by side, stacked in portrait.
- Range B spokes were decoded but discarded: the Target Analyzer confirmation only ever arrives for the range the controlling client addressed, so range B's emission gate never lifted. TA is firmware-shared between ranges, so one confirmation now lifts both gates.
- A radar whose control TCP session dropped was shown as OFF while its spokes were streaming in; spoke traffic now counts as liveness for the silence watchdog.
- Furuno radars keep one control session per client IP: a second mayara instance on the same host steals the slot back and forth with the first, controls fail on both while the picture keeps painting. This is now documented in the setup guide, and the reconnect loop detects the churn pattern and names the cause. Reconnects also back off exponentially instead of hammering (which made the fight worse), and session drops are logged with their reason instead of surfacing as a misleading "invalid range" error.
- The GUI range buttons now work before the radar has reported a range, sending the shortest supported range — which doubles as the dual-range activation command on Furuno NXT.

Known follow-up, tracked in #54: distinguish shared vs per-range controls in the UI (gain/sea/rain are firmware-shared on NXT; range, pulse width, ATF, MBS are per-range).

First working milestone of #54.

## Summary

Adds dual-range radar support across the API and GUI. The API exposes optional `dual` and `dualGroup` fields, and the GUI displays paired ranges side by side in landscape mode or stacked in portrait mode.

Improves Furuno radar handling by fixing Range B activation, synchronizing dual-range Target Analyzer state, tracking spoke traffic as liveness, adding reconnect backoff, and improving connection error reporting.

Initializes range controls when no range is reported and documents Furuno’s one-control-session-per-host limitation.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -42,6 +42,20 @@ mayara-server --openapi
 | DELETE | `/signalk/v2/api/vessels/self/radars/{id}/targets/{tid}`     | Delete tracked target                              |
 | GET    | `/signalk/v2/api/vessels/self/radars/resources/openapi.json` | OpenAPI specification                              |
 
+#### Dual-range radars
+
+A dual-range antenna (Furuno NXT, Navico 4G/HALO) appears in the radar list as
+two entries. Two optional fields on each entry identify that pairing:
+
+| Field       | Description                                                              |
+| ----------- | ------------------------------------------------------------------------ |
+| `dual`      | Which range of the antenna this entry serves — `"A"` or `"B"`             |
+| `dualGroup` | Shared key identifying the physical antenna; equal for both entries       |
+
+Both are omitted for single-range radars, so their absence means "not dual
+range" rather than "unknown". They are mayara extensions to the Signal K Radar
+API, listed as such in the API-parity allowlist.
+
 ### WebSocket Streams
 
 | Endpoint                                          | Description                                    |
--- a/docs/furuno-setup.md
+++ b/docs/furuno-setup.md
@@ -82,5 +82,16 @@ Unrecognized part codes still work with default capabilities. Please report the 
 **FAR shows "Unknown" model:**
 The part code is not in the lookup table. The radar will work with default capabilities. Report the part code from the log output so it can be added.
 
+**Controls stop working / radar flips between picture and OFF:**
+Furuno radars keep only **one control session per client IP address**. A second
+Mayara instance on the same machine (for example a containerized instance with
+host networking next to a development instance) fights the first one for that
+single slot: the radar serves each new login briefly and then closes it, both
+instances reconnect forever, and neither can hold the session. Spoke data is
+unaffected (it arrives via multicast), so the picture keeps painting while
+controls fail. Run only one Mayara instance per host when a Furuno radar is in
+use. Clients on other machines (TimeZero, MFDs) have their own IP and do not
+conflict.
+
 **Capturing traffic:**
 - For packet capture instructions used in bug reports, see [Capturing Radar Traffic](./capturing-traffic.md).
--- a/src/bin/mayara-server/web/signalk/v2.rs
+++ b/src/bin/mayara-server/web/signalk/v2.rs
@@ -207,6 +207,17 @@ struct RadarApiV3 {
     /// path but they have no effect on the recorded data stream.
     #[schema(example = false)]
     replay: bool,
+    /// Which range of a dual-range antenna this radar serves ("A", "B", ...).
+    /// Absent for single-range radars.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(example = "A")]
+    dual: Option<String>,
+    /// Group id shared by all ranges of one dual-range antenna. Clients can
+    /// pair radars carrying the same value into a combined side-by-side view.
+    /// Absent for single-range radars.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(example = "nav2452")]
+    dual_group: Option<String>,
 }
 
 impl From<&RadarInfo> for RadarApiV3 {
@@ -217,6 +228,8 @@ impl From<&RadarInfo> for RadarApiV3 {
             model: info.controls.model_name(),
             radar_ip_address: *info.addr.ip(),
             replay: info.replay(),
+            dual: info.dual.clone(),
+            dual_group: info.dual.is_some().then(|| info.base_key().to_string()),
         }
     }
 }
@@ -1632,6 +1645,20 @@ mod tests {
             model: None,
             radar_ip_address: Ipv4Addr::new(10, 56, 0, 24),
             replay: false,
+            dual: None,
+            dual_group: None,
+        }
+    }
+
+    fn radar_api(dual: Option<&str>, dual_group: Option<&str>) -> RadarApiV3 {
+        RadarApiV3 {
+            name: "4G A".to_string(),
+            brand: "Navico".to_string(),
+            model: Some("4G".to_string()),
+            radar_ip_address: Ipv4Addr::new(10, 0, 0, 1),
+            replay: false,
+            dual: dual.map(String::from),
+            dual_group: dual_group.map(String::from),
         }
     }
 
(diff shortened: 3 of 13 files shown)


dirkwa:
@keesverruijt 

I have build this "on top" of 3.4.0 API, so we don't need additional changes afterwards. 
Once 3.4.0 is merged I will change this PR against main and let CR review again.

dirkwa:
Rebased onto latest `main` — 24 commits down to 7, now mergeable.

Most of this branch had already landed on `main` via #467 (`v3.4.0 radar API`), which squash-merged the same work. A straight rebase replayed all 24 commits and conflicted immediately, with `main` holding the *newer* version of the code being reapplied — e.g. `ais_context()` in `stream.rs`. Continuing would have meant ~16 hand-resolved conflicts, each risking a revert of already-merged work.

Instead the seven genuinely-unmerged commits were cherry-picked onto current `main`. Verified as already upstream before dropping the rest: `ais_context()`, `RadarApiV3` and the `{version, radars}` envelope, `spokeDataUrl`/`streamUrl` removal, `api_surface_parity.rs`, `RADAR_API_VERSION`, and the AIS timeout handling (#499).

What remains is exactly what the description covers: the side-by-side dual view, the Furuno fixes (control-session backoff, Range B unblocking on Target Analyzer confirm, spoke traffic as liveness, the one-session-per-host doc), the control-failure cause reporting, and the range-stepping fix.

Two conflicts needed resolving, both genuine — `main` and this branch each added something that has to survive:

- `v2.rs` — `main`'s `radars_are_listed_in_a_stable_order` test (#497) alongside the branch's `dual`/`dualGroup` tests. Kept both; `radar()` gained the two new fields as `None`.
- `radar/mod.rs` — `main`'s `discriminator()` next to the branch's `base_key()`. Both kept.

All three tests pass. `cargo fmt --check` clean, `cargo clippy --all-targets -D warnings` clean, `cargo test` 409 passed / 0 failed.

dirkwa:
Addressed the review. Four applied, one declined with reasoning.

**Applied**

- **`dual.js` validation** — the ids come from the URL, so they are now checked against the radar list: both must exist and share one non-empty `dualGroup`. Stale or hand-edited links get an explicit error instead of panes that silently retry failed capability requests. I did not require the group to have exactly two members: a radar can legitimately be mid-discovery, and equal non-empty groups already expresses "two ranges of one antenna" given `dualGroup` is only set when `dual` is.
- **Furuno test coverage** — the reconnect policy is extracted into `next_reconnect_backoff()` so escalation, the cap, and the reset after a stable session are testable without a radar. Three tests added. Extracting it surfaced a real bug: the backoff was doubled twice per cycle, once after the session ended and again at the end of the loop, so the delay grew at 4× the intended rate.
- **Parity allowlist comment** — dropped the "once signalk-server adopts them" future-migration note.
- **`dual.css`** — replaced the echo comment with why each pane is an iframe.

**Declined — bumping `api-version` to 3.5.0**

I made this change first, and the API parity test rejected it. The shared surface is unchanged: `fields=brand,model,name,radarIpAddress` before and after, because `dual`/`dualGroup` are excluded as documented mayara extensions. The version pins what *both* projects implement, so bumping it unilaterally diverges mayara from signalk-server (the test says to bump "in lockstep"), and the GUI's `^major.minor` check would then reject a 3.4.0 server over fields it never needed.

The documentation half of that finding was worth doing regardless, so `dual`/`dualGroup` are now described in `docs/api/README.md`.

`cargo fmt --check` clean, `cargo clippy --all-targets -D warnings` clean, `cargo test` 388 passed / 0 failed.

dirkwa:
Second round addressed (found by a local `cr` run — the CodeRabbit check on this PR is green but has not re-reviewed since 02:48Z, before the fix commit).

- **Doc comment on `next_reconnect_backoff`** — trimmed the two lines that restated the signature, keeping only why it is split out.
- **Untested `reached_data_loop` path** — valid gap, and fixing it surfaced an ordering bug. A setup failure (socket, login, or command stream) escalated the backoff *after* the sleep, so the first retry always waited the old delay. That path now goes through `next_reconnect_backoff()` with a zero-length session, which both fixes the ordering and makes it testable. Test added covering escalation and the cap.

`cargo fmt --check` clean, `cargo clippy --all-targets -D warnings` clean, `cargo test` 413 passed / 0 failed.

keesverruijt (commented):
@dirkwa — tested this on a HALO24 (dual range, `nav1034A`/`nav1034B`) and the side-by-side view works. One note on `1016264` and an idea for the Furuno side that needs a DRS to confirm.

## `1016264` is fine to keep, but it is solving the wrong problem for Navico

The guard fires when the reported range is not in `validValues`, and on Navico that is not an activation quirk — it is the normal state. With the chart overlay driving the range, the radar sweeps whatever suits the chart. Measured here across one session:

```
2700 m (1.45 nm)   4517 m (2.43 nm)   6543 m (3.53 nm)   9888 m (5.33 nm)
12645 m (6.82 nm)  16308 m (8.80 nm)  21157 m (11.42 nm) 35918 m (19.39 nm)
```

Two of 22 distinct values were in the range table. The MFD displayed "2.5 nm" at one point, which is not selectable on that radar at all, and the raw values even jitter (65433 and 65434 decimetres for one nominal setting).

So on Navico the guard would fire in ordinary use and send `validValues[0]` — dropping the operator from 5 nm to 57 m on a single zoom press.

**#515** fixes that at the source: a numeric control snaps a value to the nearest settable one, in `Control::set`, so it applies both to what the radar reports and to what a client asks for. Enums are excluded, because `Power` lists `[1, 2]` for Standby and Transmit and rounding a stray Fault to the nearest would start the radar. That is verified on the water — the same radar that was reporting 9888 m now reports 11112 m. We will rebase #515 onto main and merge it.

With that in, `currentIndex === -1` should not occur on Navico any more, so `1016264` becomes belt-and-braces rather than load-bearing. No need to change it here.

## What we think the Furuno path should do instead

The Furuno case is genuinely different: Range B has **no range at all** until its first command activates the channel. That is an absent value, not an off-table one, so #515 does not help — there is nothing to snap.

One correction to an assumption we started with: mayara does not need to ask the radar for its valid ranges. They are model-derived already, in `furuno/settings.rs`:

```rust
let ranges = Ranges::new_by_distance(&get_ranges_by_model(&model));
```

So `validValues` is populated for Range B as soon as the model is known. What is missing is the **current** value.

The protocol can be asked for it. `CommandMode::Request` (`'R'`) exists and mayara already uses it for `Modules`, `OnTime`, `TxTime` and `Status`; `CommandId::Range` is `0x62`, so `$R62` is available and nothing sends it today.

The shape we would suggest:

- when Range B is detected in use — being controlled, or spokes arriving for it — and its range is unknown,
- send `$R62` for that channel and publish whatever comes back,
- so the GUI needs no special case: there is a value to step from and `zoomIn`/`zoomOut` work normally.

**The open question, and what would be useful to test on a DRS:** what does a Furuno reply with for a channel that has not been activated yet? It might report the range it would use, or nothing, or an error. If it answers, this replaces the GUI fallback with something correct for every client. If it does not, we are back to needing a default — but then it belongs in the Furuno backend where it can be brand-specific, rather than shipped to every client in `control.js`.

We did not build it, because it is speculative protocol behaviour without a radar to answer that question.

dirkwa:
Thanks — the Navico measurement is convincing, and I agree `1016264` is solving the wrong problem there. Two things from testing the Furuno side on a live DRS4D-NXT.

## `$R62` is already sent

The premise that nothing sends it today is not quite right — `command.rs` `init()` already issues it:

```rust
self.send(CommandMode::Request, CommandId::Range, &[]).await?; // $R62
```

It goes out with no `drid` argument, so it queries Range A only (the comment above that block says "Range A"). Adding the channel would be `$R62,1`, since `dual_range_id` is already how every other per-range command is addressed.

## What the radar answers for an unactivated Range B

This is testable, so I tested it rather than leaving it open. Both channels on a DRS4D-NXT, same moment:

| channel | reported range | timestamp |
| --- | --- | --- |
| `fur6424A` | 463 | live, updates as the range changes |
| `fur6424B` | **0** | frozen at radar startup, never updated since |

So the radar **does** answer for a channel that has never been activated, and the answer is `0`. `validValues` for Range B is fully populated (`[115, 231, 463, 926, …]`) and does not contain `0`, which matches your point that the ranges are model-derived and available before activation.

That means asking `$R62,1` would not help: the value comes back, it is just not a usable range. It is an absent value dressed as a number — the "or nothing" branch of your question, in practice.

## Where that leaves the GUI guard

Given the above I would keep a fallback, but I take your point that shipping it to every client in `control.js` is the wrong home for something brand-specific — especially now that `currentIndex === -1` should not occur on Navico once #515 lands.

The cleaner shape, if you agree: have the Furuno backend treat a Range B range of `0` as "not yet activated" and publish no value at all, rather than `0`. Then the GUI's existing "no value" path is correct for every brand, the fallback lives in the backend where the `0` convention is known, and `control.js` needs no special case. Happy to do that here or in a follow-up — your call on scope, since this PR is already carrying the dual-range view plus the Furuno fixes.

Also glad the side-by-side view works on a HALO24 — that is the first confirmation outside a Furuno.

PR #483, opened 2026-08-04
feat(network): advertise web server via mDNS

keesverruijt:
Mayara is now discoverable on the local network. From any computer, tablet or phone on the same network the web interface can be opened at `http://mayara.local:6502` — no need to look up the server's IP address first — and network browsers (Bonjour/Avahi, Signal K servers, other tooling) can find mayara automatically.

## Details

Registers the DNS-SD service `_mayara-http._tcp.local.` under hostname `mayara.local.`, using the `mdns-sd` crate that was already a dependency (until now only used for browsing).

- TXT records: `version` (mayara), `api` (Signal K radar API), `path=/signalk`, `scheme` (`http`/`https`, following `--tls-cert`/`--tls-key`).
- `enable_addr_auto()` keeps the address records in step as interfaces and IPs come and go.
- The daemon probes both names, so a second mayara on the same LAN is renamed rather than clashing.
- Registration happens after the listener binds, so a port that is already in use is never announced. Any mDNS error is logged as a warning and the web server continues.
- Dropping the advertiser unregisters (waiting briefly for the goodbye packets) and shuts the daemon down, so clients drop the service immediately instead of waiting out the TTL.

There is no flag to disable the announcement, and it is not scoped to `--interface` — that flag limits radar discovery, whereas the web server listens on all interfaces.

## Test plan

- [x] `cargo test` — 355 lib tests + integration tests pass, including two new unit tests over the generated `ServiceInfo`
- [x] Live run on macOS: `dns-sd -B _mayara-http._tcp` finds `mayara`; `dns-sd -L` resolves `mayara.local.:6599` with the expected TXT records; `curl http://mayara.local:6599/signalk` returns the endpoints document; Ctrl-C makes the service disappear from the browser immediately

Adds always-enabled mDNS/DNS-SD advertising for the Mayara web server.

- Registers `_mayara-http._tcp.local.` under `mayara.local.` after the web listener binds.
- Advertises the Mayara version, Signal K radar API, `/signalk` path, HTTP/HTTPS scheme, and port.
- Updates address records when network interfaces change.
- Handles hostname conflicts through probing and renaming.
- Logs advertising errors without stopping the web server.
- Unregisters the service and shuts down mDNS when the advertiser is dropped.
- Documents discovery commands for macOS and Linux.
- Adds unit tests for service identity, addresses, port, and TXT properties.
- Passes tests, formatting, clippy, and live macOS discovery checks.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -198,6 +198,15 @@ mayara-server --nmea0183 -n udp:0.0.0.0:10110
 
 The built-in web interface is available at `http://localhost:6502` (or your configured port).
 
+Mayara announces itself on the local network with mDNS (Bonjour/Avahi), so from
+another computer, tablet or phone on the same network the web interface is
+reachable at `http://mayara.local:6502` without knowing the server's IP address.
+The announcement is the DNS-SD service type `_mayara-http._tcp`, with TXT keys
+`version`, `api` (Signal K radar API version), `path` and `scheme`
+(`http` or `https`, depending on whether TLS is configured). Browse for other
+mayara servers with, for example, `dns-sd -B _mayara-http._tcp` (macOS) or
+`avahi-browse -r _mayara-http._tcp` (Linux).
+
 Features:
 - Real-time radar display (PPI view)
 - Range and control adjustments
--- a/src/bin/mayara-server/web/mod.rs
+++ b/src/bin/mayara-server/web/mod.rs
@@ -161,6 +161,19 @@ impl Web {
         socket.listen(1024).map_err(WebError::Io)?;
         let listener = TcpListener::from_std(socket.into()).map_err(WebError::Io)?;
 
+        // Only announced once the port is actually ours. Discovery is a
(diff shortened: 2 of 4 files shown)


keesverruijt:
CodeRabbit findings triaged:

1. **`--port 0` advertised as port 0** — valid, fixed in 0fa832a. The advertisement now uses `listener.local_addr().port()`, so an ephemeral port is announced correctly.
2. **Docs should not imply a fixed URL** — skipping. The module doc already writes the port as `<port>`, and the USAGE.md section states "(or your configured port)" and lists the `scheme` TXT key with both `http` and `https`; `http://mayara.local:6502` is presented as the default-configuration example.
3. **Consume `daemon.monitor()` to log `DaemonEvent::Error`** — skipping. It needs a spawned task plus shutdown plumbing to surface errors the web server would not act on anyway; registration failures are already logged and the daemon resolves name conflicts by itself. Worth revisiting if conflict renames turn out to confuse users in the field.

PR #489, opened 2026-08-04
fix(web): make /quit stop the whole server

keesverruijt:
Asking mayara to quit left it running. `GET /quit` answered `bye` and the web interface went away, but the process stayed alive — still consuming radar and Signal K traffic, no longer reachable, and only stoppable by killing it by hand. Worse, it had released the HTTP port, so starting mayara again succeeded and left two instances competing for the same radar. The only clean way to stop mayara was Ctrl-C or a signal from the service manager.

Now `/quit` shuts the whole process down, the same way Ctrl-C does.

## Details

`quit_handler` sends on `Web::shutdown_tx`, a web-local broadcast channel whose only subscribers are the axum graceful-shutdown future and the outbound WebSocket streams. Nothing connected it to tokio-graceful-shutdown: `axum::serve` returned `Ok`, `Web::run` returned `Ok(())`, and a subsystem finishing normally is not a shutdown request. The toplevel cancellation token was never cancelled, so Locator, NavData, Radar Watchdog, AIS Broadcast/Timeout/Seed, TrackerManager and the per-radar subsystems all kept waiting on `on_shutdown_requested()`, and `handle_shutdown_requests` — which returns once a shutdown is requested *or* all subsystems have finished — waited forever.

The fix calls `subsys.request_shutdown()` where the serve future completes, cancelling the toplevel token and unwinding the tree through the path signals already use. It is placed before the error return, so a web server that dies of an I/O error also brings the process down instead of leaving the same headless state.

## Test plan

- [x] New `tests/quit.rs` starts the binary on an ephemeral port, hits `/quit`, and asserts the process exits — verified it **fails** without the fix ("server still running after /quit") and passes with it
- [x] Live: `curl /quit` against a normally started server now exits in ~2 s; before the fix the process was still alive with 20 threads and needed SIGINT
- [x] `cargo test --features pcap-replay` — 367 lib tests plus all integration tests pass
- [x] `cargo fmt --check` and `cargo clippy --no-deps --all-targets --all-features -- -D warnings` clean

The test runs the server with `--parent <test pid>`, so it binds loopback only and does not advertise on mDNS — a test run neither exposes a port nor spams the LAN, and an orphaned server dies with the test process.

Fixes `/quit` to shut down the complete Mayara process. When the web server stops or reports an I/O error, it requests subsystem shutdown and cancels the toplevel token.

Adds an integration test that starts `mayara-server`, requests `/quit`, and verifies that the process exits successfully.

### DIFF
--- a/src/bin/mayara-server/web/mod.rs
+++ b/src/bin/mayara-server/web/mod.rs
@@ -257,6 +257,9 @@ impl Web {
                     let _ = shutdown_tx.send(());
                 },
                 r = axum::serve(tls_listener, app).with_graceful_shutdown(shutdown) => {
+                    // Reached only via /quit or a serve error; either way the
+                    // rest of mayara must not keep running headless.
+                    subsys.request_shutdown();
                     return r.map_err(WebError::Io);
                 }
             }
@@ -272,6 +275,9 @@ impl Web {
                     let _ = shutdown_tx.send(());
                 },
                 r = axum::serve(listener, app).with_graceful_shutdown(shutdown) => {
+                    // Reached only via /quit or a serve error; either way the
+                    // rest of mayara must not keep running headless.
+                    subsys.request_shutdown();
                     return r.map_err(WebError::Io);
                 }
             }
--- a/tests/quit.rs
+++ b/tests/quit.rs
@@ -1,0 +1,80 @@
+//! `/quit` must stop the whole process, not just the web server.
+//!
+//! Regression test: the web server used to shut down on its own broadcast
+//! channel without ever requesting a shutdown of the subsystem tree, leaving
+//! a headless process behind that still held the radar sockets and could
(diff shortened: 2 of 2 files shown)


keesverruijt:
CodeRabbit findings triaged:

1. **Historical narrative in the test module doc** — valid, fixed. AGENTS.md asks for documentation of current state, so the doc now states the requirement and the failure mode it guards, in present tense.
2. **Unnamed 100 ms poll interval** — valid, fixed. Now `EXIT_POLL_INTERVAL`, per the no-magic-numbers rule.
3. **No test for the TLS serve branch** — skipping. Both arms of the select got the identical two-line change, and covering the TLS branch means generating a certificate and key in the test harness, which is disproportionate to the risk of one duplicated line. Worth doing if TLS-specific shutdown logic ever appears.

PR #493, opened 2026-08-04
feat(output): show any radar brand on a Garmin chartplotter

keesverruijt:
A Garmin GPSMAP only draws radar from a Garmin radar. Built with the optional
`garmin-xhd-output` feature, Mayara now emulates a GMR xHD on the Garmin Marine
Network and feeds it the picture of whichever radar it found — Furuno, Navico,
Raymarine, Koden, or the built-in emulator. The plotter discovers it by itself
and lists it as a GMR xHD.

Controls work both ways: range, gain, sea clutter, rain clutter and the
transmit button on the plotter change the setting on the real radar, and a
change made from Mayara's GUI shows up on the plotter.

The feature is off by default; it needs `cargo build --features garmin-xhd-output`
and a `172.16.x.x` address on the interface the plotter is reached through.
See [docs/garmin-xhd-output.md](docs/garmin-xhd-output.md).

Supersedes #478 by @matztam, whose work this builds on — closes #478.

## What changed against #478

The bridge was Furuno-only. It now hangs off `RadarInfo::start_outputs()`,
which replaces `start_forwarding_radar_messages_to_stdout()` and is called by
every brand, so nothing in `src/lib/output/` knows which brand produced the
spokes it converts. Three things turned out to be brand-specific rather than
one:

- **Sample values** are legend indices, not echo strength. They are mapped
  through a table built from the radar's own `Legend`; previously a 16-level
  Navico or a 2-level HD radome would have drawn a nearly black picture.
  Doppler classes, which an xHD has no room for, become plain strong echoes.
- **Spoke counts** differ by a factor of thirty across brands. Spokes are
  merged or repeated onto the xHD's 1440 per revolution, which fills the
  picture of a 250-spoke Quantum and stops an 8192-spoke Furuno from sending
  five times the traffic a real xHD does.
- **Range**: the spoke header's `range_meters` is the Range control and
  `display_meters` is `spoke.range`, which is what those two fields mean. The
  hardcoded 1.7822 Furuno correction factor is gone.

Also: bursts are drained at the rate they arrive rather than a fixed 200 µs;
commands go through `SharedControls::process_client_request()` so they are
validated and unit-converted like any other client request, with rejections
logged; acknowledgements leave from the report socket the plotter expects them
on; and the capability bitmap is constructed from named bits listing exactly
what the bridge honours, instead of replaying a capture that claims dual range
the bridge does not have.

One radar is bridged per instance, since the emulated radar owns fixed
multicast groups and ports. Multicast loopback stays on so a display on the
same host sees it too; Mayara avoids discovering its own emulation by ignoring
beacons from the single address it announces from.

## Testing

31 unit tests cover the spoke conversion (angle mapping across six source spoke
counts, peak-preserving resampling, legend and Doppler mapping, the two range
fields), the command translation, the report stream, and the capability set.
CI now runs clippy and the test suite with the feature enabled.

Verified end to end on loopback, with a second Mayara standing in for the
plotter (recipe in the docs): the emulated radar is discovered, reports
`dual_range=false`, registers as a single radar, offers exactly the five
controls the bridge translates, and streams spokes at 577/s — 1440 per
revolution at 24 RPM.

Not verified on a real Garmin chartplotter — I have no hardware. The two
behaviours most worth watching there are the reduced capability set and the
range acknowledgement, which now holds the plotter's requested range until the
radar confirms it instead of locking for a fixed 5 s.

## Summary

- Adds the optional `garmin-xhd-output` feature.
- Emulates a Garmin GMR xHD radar for Garmin chartplotters.
- Converts and streams data from supported radar brands over the Garmin Marine Network.
- Forwards range, gain, sea clutter, rain clutter, and transmit controls.
- Adds discovery, status, spoke, command, acknowledgement, and range handling.
- Integrates output startup through `RadarInfo::start_outputs()`.
- Adds network validation, documentation, 35 bridge tests, and feature-enabled CI checks.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,6 +37,8 @@ jobs:
       run: cargo clippy --no-deps --all-targets -- -D warnings
     - name: cargo clippy (--features pcap-replay)
       run: cargo clippy --no-deps --all-targets --features pcap-replay -- -D warnings
+    - name: cargo clippy (--features garmin-xhd-output)
+      run: cargo clippy --no-deps --all-targets --features garmin-xhd-output -- -D warnings
 
   build:
     name: Build ${{ matrix.target }}
@@ -74,7 +76,7 @@ jobs:
     - uses: actions/checkout@v7
     - uses: swatinem/rust-cache@v2
     - name: Run tests
-      run: cargo test --features pcap-replay --release
+      run: cargo test --features pcap-replay,garmin-xhd-output --release
     - name: Build
       run: cargo build --release
     - name: Upload artifacts
@@ -93,7 +95,7 @@ jobs:
       with:
         tool: nextest
     - name: Run tests
-      run: cargo nextest run --features pcap-replay
+      run: cargo nextest run --features pcap-replay,garmin-xhd-output
     - name: Build
       run: cargo build --release
     - name: Upload artifacts
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,9 @@ koden = []
 raymarine = []
 emulator = []
 pcap-replay = []
+# Emulate a Garmin GMR xHD radar so a Garmin chartplotter can display any
+# radar mayara has found. Built on the `garmin` brand's protocol definitions.
+garmin-xhd-output = ["garmin"]
 # default = ["navico", "furuno", "raymarine"]
 default = ["navico", "furuno", "garmin", "koden", "raymarine", "emulator"]
 
--- a/README.md
+++ b/README.md
@@ -42,6 +42,10 @@ Implemented but awaiting real-hardware validation:
 
 If your radar is not on the fully supported list, contact us to help with testing!
 
+Mayara can also work the other way around: an optional build feature makes it
+[emulate a Garmin radar](docs/garmin-xhd-output.md), so a Garmin chartplotter
+can display the picture of a radar of any other supported brand.
+
 ## End users
 
 Mayara runs on Windows, macOS, and Linux (including Raspberry Pi). Download a pre-built binary, start it, and open a web browser — no programming required.
--- a/docs/garmin-xhd-output.md
+++ b/docs/garmin-xhd-output.md
@@ -1,0 +1,141 @@
+# Showing Any Radar on a Garmin Chartplotter
+
+A Garmin GPSMAP chartplotter only draws radar from a Garmin radar. Mayara can
+close that gap: built with the `garmin-xhd-output` feature it emulates a Garmin
+GMR xHD on the Garmin Marine Network, and feeds it the picture of whichever
+radar it found — a Furuno DRS, a Navico HALO, a Raymarine Quantum, a Koden, or
+the built-in emulator. The plotter discovers it by itself and shows it in the
+radar list as a GMR xHD.
+
+The controls work in both directions. Range, gain, sea clutter, rain clutter,
+and the transmit button on the plotter change the setting on the real radar,
+and a change made from Mayara's own GUI or another display shows up on the
+plotter.
+
+> **Not yet validated on a chartplotter.** The bridge has been verified against
+> Mayara's own Garmin receiver over loopback — discovery, the report stream,
+> controls and spoke data all check out — but nobody has yet put it in front of
+> a physical Garmin GPSMAP. If you try it, please report how it goes.
+
+## What you need
+
+- Mayara built with the feature (it is not in the standard binaries):
+
+  ```sh
+  cargo build --release --features garmin-xhd-output
+  ```
+
+- An address on the Garmin Marine Network, `172.16.x.x` with netmask
+  `255.240.0.0`, on the interface the plotter is reachable through. If the
+  radar is on a different subnet — a Furuno on `172.31.x.x`, say — give the
+  same interface both addresses:
+
+  ```text
+  enp0s13f0u2
+    172.31.254.217/16   Furuno radar
+    172.16.254.217/12   Garmin Marine Network
+  ```
+
+  Mayara picks the Garmin address on the radar's own interface. When the
+  plotter is on an interface of its own, any `172.16.x.x` address will do;
+  addresses in `172.17`–`172.31` are skipped, because that is where Docker and
+  similar bridges live rather than Garmin equipment.
+
+- Mayara and the plotter on the same network segment. Multicast is sent with a
+  TTL of 1, so nothing crosses a router.
+
+The log says which radar is being emulated and from where:
+
+```text
+Furuno-1234: Garmin xHD output emulating a GMR xHD on 172.16.254.217
+```
+
+## One radar at a time
+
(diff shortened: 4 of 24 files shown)


keesverruijt:
Thanks @coderabbitai — addressed in a9559754. Eleven of the thirteen applied; two declined with reasons below.

**Fixed**

- **`convert.rs` spoke index** — real bug, and the best catch of the review. The modulo ran after the `as u16`, so an angle past a full revolution was truncated onto an unrelated spoke instead of wrapping. Now wrapped in `u64` first, with a regression test: angle 100000 at 2048 spokes/rev lands on spoke 1192, where narrowing first gave 456.
- **`command.rs` sea mode** — an unrecognised mode fell into the wildcard and was read as "switch sea clutter off", quietly changing the picture. Unknown modes are now logged and dropped. Applying the fix surfaced a second bug the compiler caught: `SEA_MODE_OFF` was unimported, so it bound as a catch-all pattern rather than matching the constant. Both covered by tests.
- **`command.rs` payload length** — checked addition; a bogus length can no longer overflow on a 32-bit target.
- **`cdm.rs` and `status.rs` literals** — named throughout (version marker, service fields, tail tag/length; antenna power, trigger period/delay, tune and status values, input voltage as decivolts).
- **`capabilities.rs` prefix provenance** — now cites the pcap fixture, the offset, and the MFD function (`radar_parse_xhd_capability_packet`, `FUN_0490ef30`) that reads the five words at fixed offsets and never looks at the prefix.
- **`spokes.rs` tests** — added for the three module-boundary behaviours: malformed batch, spokes with `range == 0`, and a queue trimmed past `QUEUE_MAX`.
- **`radar/mod.rs`** — `start_outputs` narrowed to `pub(crate)`; nothing outside the crate calls it.
- **Docs** — fences tagged `text`, README no longer implies a Garmin radar can be bridged to a Garmin plotter, and the setup guide now carries an explicit note that this is unvalidated on physical hardware.

**Declined**

- **`status.rs:208`, use the shared spoke-count constant for `MSG_SPOKE_TOTAL`** — that would change 2400 to 1440 and make the emulation diverge from real hardware. They are different quantities: a captured GMR xHD reports `spoke_total = 0x0960 = 2400` (`research/garmin/discovery-handshake.md:185`) while sending 1440 spokes per revolution. Kept the value, but named it `SPOKE_TOTAL` with that provenance so it is no longer a bare literal.
- **`output/mod.rs:4`, gate `garmin_xhd` behind the feature** — already gated one level up: `src/lib/mod.rs:25` declares `pub(crate) mod output;` under `#[cfg(feature = "garmin-xhd-output")]`, so neither the module nor its children compile without it. A second gate would be redundant.

Verified after the changes: clippy clean with and without the feature, 35 bridge tests plus the full suite passing, and the loopback end-to-end check still discovers a single radar reporting `dual_range=false`.

matztam:
I tested this PR with a Furuno DRS4D-NXT (low-power model) bridged to a Garmin GPSMAP chartplotter and found three bugs. All fixes are in my fork at [matztam/mayara-server, branch feat/garmin-xhd-output-pr493](https://github.com/matztam/mayara-server/tree/feat/garmin-xhd-output-pr493). Sharing the details here in case they are useful for the upstream PR.

---

## Bug 1: Sea/Rain controls not visible on the Garmin plotter

The capability bitmap in `status.rs` was hand-picked and too sparse — even though the correct bits for sea and rain were present, the Garmin MFD requires a much fuller capability footprint before it shows those controls. Switching to the verbatim bytes from a real GMR xHD capture (passed through `GarminCapabilities::parse` / `to_body()` so the struct stays the authority) fixed this. Gain, Sea and Rain controls all appeared immediately.

## Bug 2: All radar echoes solid red, no colour gradation

The Furuno DRS in low-power mode applies an 18th-root gamma curve when mapping raw wire bytes (0–252) to legend indices. With `ECHO_FLOOR = 10` and `pixel_colors = 252`, almost all wire bytes land in legend indices ~180–251 — the top 28% of the scale. The original `intensity_table()` scaled linearly from index 0, so every real echo from the radar mapped to xHD strength 182–255 and appeared solid red.

**Fix** — two parts:

1. Added `wire_gamma: Option<f64>` to `Legend`. Furuno's `parse_modules()` sets `wire_gamma = Some(18.0)` and `low_return = ECHO_FLOOR` when a low-power model is identified.

2. `intensity_table()` now scales from `low_return` (not 0) and, when `wire_gamma` is set, applies the inverse gamma (`normalized.powf(g)`) before mapping to 0–255. This undoes the radar's compression and restores the full dynamic range: faint returns appear yellow/green, strong ones red.

```rust
let normalized = (index - floor) as f64 / span as f64;
let corrected = match legend.wire_gamma {
    Some(g) => normalized.powf(g),
    None => normalized,
};
table[index] = (corrected * u8::MAX as f64).round() as u8;
```

## Bug 3: Range changes from the Garmin have no effect

The Garmin sends range values that are off by up to 1 m from the xHD table entries it was offered (e.g. 2777 instead of 2778, 3703 instead of 3704). The Furuno rejects any value not exactly in its own range table, so every range change silently failed.

The original code already called `nearest_xhd_range()` for `set_pending_range()`, but forwarded the raw (unsnapped) value to the radar via `ControlValue`. Applying `nearest_xhd_range()` to both fixed it:

```rust
MSG_RANGE_A => long.map(|meters| {
    let snapped = nearest_xhd_range(meters);
    shared.set_pending_range(snapped);
    ControlValue::new(ControlId::Range, json!(snapped))
}),
```

---

## Bonus: Rain-off → mayara Auto

Minor UX improvement: pressing "Rain off" on the Garmin now engages mayara's automatic rain clutter mode rather than forcing the level to 0%. A hard 0% is still reachable via the gain slider. The status report was also fixed to send `rain_mode=0` when mayara is in auto (previously it echoed the underlying gain percentage, leaving the Garmin stuck showing e.g. 50%).

keesverruijt:
I don't have a lot of time today to respond, but thank you for your efforts and would like to give some quick feedback.

1) Sea/Rain controls not visible.

This is problematic for me. Just dumping all capabilities as if it is a real xHD is a kludge, not a real fix. I will share some background in DM.

2) Furuno is indeed special in what it puts in its spokes. I will confer more with @dirkwa to see what we can do. IMO mayara clients (including the "forward to garmin" code) should be able to depend on the legend as if it is linear.

3) This should actually be fixed by #515 , can you verify that your fix is maybe not needed anymore if you rebase onto main?

Bonus) great!

matztam:
> 3) This should actually be fixed by #515, can you verify that your fix is maybe not needed anymore if you rebase onto main?

Confirmed — tested on the water after rebasing onto main with #515 included. Range changes from the Garmin plotter work correctly without our `command.rs` fix. #515 handles it at the right layer (snapping against the radar's own `valid_values` in `process_client_request`), so we dropped our xHD-specific workaround.

keesverruijt:
I'd rather wait with this another 2 months or so, when I will have access to some hardware to run this myself.

PR #502, opened 2026-08-08
fix(targets): alarm guard zones on target promotion, not first blob

keesverruijt:
A vessel entering an active guard zone usually raised no alarm. Signal K clients saw no `notifications.radar.<id>.guardZone.<n>` delta, and once a zone had gone quiet it often stayed quiet for the rest of the session — even as targets kept being acquired inside it. The GUI's own audible alert kept working throughout, which made the silence look like a client problem rather than a server one.

Three things went wrong, all in how the alarm was driven.

## Cause

**It fired on acquisition, not on a confirmed target.** The alarm was raised on `ProcessResult::NewAcquiring` — the first speculative blob inside the zone. A vessel already being tracked when it crossed into the zone matched an existing target instead, yielding `Updated`, so the most ordinary case of all produced nothing.

**A counter suppressed repeats.** Alarms were gated on a 0→1 transition of a per-zone active count, so a second vessel entering an already-alarming zone was silent.

**The counter leaked, permanently.** It was decremented only when a target went `Lost`. `delete_target` and `clear_all_targets` removed targets without decrementing, and a target revived after `Lost` could decrement twice for a single increment. Once the count was stuck above zero, no further 0→1 transition was possible and the zone never alarmed again.

## Fix

Alarm on **promotion** — the same event the GUI's audible alert already used, and one that means the target survived four revolutions rather than being a single clutter return. Alert per target rather than only on the zone's first, so a second vessel entering an alarming zone is still announced. Derive the clear from tracker occupancy recomputed each sweep, rather than tallying enter/leave events, so the `Lost` timeout, an explicit delete and a radar-wide clear all release the alarm identically and no counter can drift.

Occupancy is defined as the same predicate that decides whether a target is broadcast at all (`MIN_BROADCAST_UPDATE_COUNT`), so the alarm and what a client can see cannot disagree.

## Testing

Seven tests, driving real `process_blob` traffic and asserting on the serialized `notifications.*` deltas a client would receive rather than on internal counters. Three fail against the old code: silence below promotion, an alert for the *second* target in a zone, and the clear after `delete_target`.

## Note for testing on real hardware

Guard zones currently collapse to a 0-0 range when edited by dragging, which leaves a zone that can never contain a target and so can never alarm. That is a separate, pre-existing bug with a fix on `fix/zone-partial-update`; it is worth having both in place before judging this one on the water.

## Summary

- Triggers guard-zone alarms when targets become promoted after sustained tracking.
- Sends one alert for each target entering a zone.
- Computes zone occupancy from client-visible tracker state.
- Clears alarms consistently after target loss, target deletion, timeout, or bulk target removal.
- Excludes manual MARPA targets from occupancy.
- Centralizes the minimum broadcast update threshold.
- Adds seven tests using real `process_blob` traffic and serialized `notifications.*` deltas.

### DIFF
--- a/src/lib/radar/target/manager.rs
+++ b/src/lib/radar/target/manager.rs
@@ -10,10 +10,17 @@ use chrono::{DateTime, Utc};
 use tokio::sync::{broadcast, mpsc};
 
 use super::blob::CompletedBlob;
-use super::tracker::{CandidateSource, ProcessResult, TargetCandidate, TargetTracker};
+use super::tracker::{
+    CandidateSource, ProcessResult, TargetCandidate, TargetStatus, TargetTracker,
+};
 use super::{ArpaTargetApi, TargetDangerApi, TargetMotionApi, TargetPositionApi};
 use crate::radar::{GeoPosition, KN_TO_MS};
-use crate::stream::SignalKDelta;
+use crate::stream::{NotificationMethod, NotificationState, NotificationValue, SignalKDelta};
+
+/// Updates a target needs before it is broadcast to clients. Below this
+/// the motion estimate is too green to be worth reporting, so a target
+/// that dies young is never seen by a client and never alarms.
+const MIN_BROADCAST_UPDATE_COUNT: u32 = 4;
 
 /// Context from the spoke that produced a blob
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


keesverruijt:
@coderabbitai review

PR #510, opened 2026-08-08
feat(radar): identify radars from their own protocol, not their address

keesverruijt:
A radar's saved settings — gain, range units, guard zones, the name you gave it — were lost whenever its IP address changed. On a boat where the radar picks up a new DHCP lease, it came back looking like a radar mayara had never seen. Two radars of the same brand could also collide on one identity and only one of them would appear.

Radars were identified by the low 16 bits of their address, because nothing better was being read off the wire. In fact every brand publishes something better, and mayara was already receiving it.

## Where each identity comes from

| brand | identity | message |
|---|---|---|
| Raymarine | `link_id` | 36/56-byte beacons |
| Garmin | unique id | `0x038E` CDM heartbeat |
| Koden | its own MAC | `0xA7` response |

All three are available at discovery, before a radar is named, and all three are stable across reboots and independent of the address. Verified against recordings: Raymarine link_ids hold across five separate captures of one unit and across a power-up; Garmin's id is unchanged across a cold boot; both arrive from two different source IPs for the same device.

Koden already asked for its MAC — the answer was logged and discarded. The request now comes from the locator, which listens on the port the answer arrives on.

The address remains the last resort for anything that publishes nothing.

## Upgrade effect

Radars that gain an identity change key once. Settings saved under the old address-derived key are adopted automatically on the first run, so nothing is lost. The old key is derived by calling the very function that produced it, so the migration cannot drift out of step with the format it is looking for.

## Also fixed

Garmin's `product_id` and its device name and alias were stored under the CDM port and looked up under the radar's data port, so they never reached the radar: the model name fell back to capability inference and a user's own alias was ignored. Keyed by address now. It is a prerequisite for the identity work above rather than a separate change, but happy to split it if you would rather review it on its own.

## Testing

Unit tests for each brand's identity parse, including truncated input, plus replay assertions against the shipped fixtures pinning the key to the identity — `ray68b4` from a link_id, `gar0aa0A` from a heartbeat id — and, for Raymarine, pinning the key apart from the real serial number that is shown to the operator.

## Note

Earlier revisions of this branch derived identity from the host's ARP cache. That is gone: it needed a wait, platform-specific table reads, and traffic to provoke resolution, and it could not work in replay. Reading each brand's own protocol removes all of it.

## Summary

Use stable protocol-provided identifiers for radar identity:

- Raymarine uses beacon `link_id`.
- Garmin uses the CDM heartbeat `unique_id`.
- Koden uses the device MAC address.
- The IP address remains the fallback.

Migrate saved settings from the previous address-derived key. Key Garmin metadata by IP address instead of CDM port. Keep radars visible while identity is unavailable. Add identity parsing, truncated-message, malformed-response, retry, throttling, migration, and replay tests.

### DIFF
--- a/src/lib/brand/garmin/discovery.rs
+++ b/src/lib/brand/garmin/discovery.rs
@@ -28,15 +28,17 @@
 //! +07     1     constant             0x01
 //! +08     1     service_count        0x01
 //! +09     3     padding              0x00 0x00 0x00
-//! +0c     8*N   service_id_array     [class:1][inst:0][ver:2][rsv:0]+u32 id
+//! +0c     4*N   service_id_array     one u32 per published service
+//!         4     unique_id            per-device, stable across power cycles
 //!         var   serialized tail      uptime/sequence counter
 //! ```
 
 #![allow(dead_code)]
 
 use super::protocol::{
-    CDM_OFFSET_PRODUCT_ID, CDM_OFFSET_PRODUCT_SUBTYPE, CDM_OFFSET_SIMULATOR_MODE,
-    CDM_OFFSET_SYC_GROUP_ID, CDM_OFFSET_VERSION_MARKER,
+    CDM_OFFSET_PRODUCT_ID, CDM_OFFSET_PRODUCT_SUBTYPE, CDM_OFFSET_SERVICE_COUNT,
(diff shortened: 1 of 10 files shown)


keesverruijt:
@coderabbitai review

keesverruijt:
@coderabbitai review

keesverruijt:
@coderabbitai review

PR #512, opened 2026-08-09
fix(api): list radars in a stable order

### LINKED ISSUE #497: GET /radars returns radar keys in nondeterministic order
The `radars` map in `GET /signalk/v2/api/vessels/self/radars` comes back in a different key order from one request to the next, because it is serialised from a hash map. Clients that take "the first radar" therefore get a different radar at random.

## Observed

Against mayara 3.8.1 with a dual-range HALO24 (`nav1034A` / `nav1034B`), six consecutive requests:

```
[nav1034B, nav1034A]
[nav1034B, nav1034A]
[nav1034B, nav1034A]
[nav1034B, nav1034A]
[nav1034B, nav1034A]
[nav1034A, nav1034B]   <-- flipped
```

`GET /radars/{id}/controls` does **not** have this problem — its keys come back alphabetically sorted every time (`accentLight`, `antennaForward`, `antennaHeight`, …), so this looks specific to the radar list.

## Why it matters

JSON objects are unordered by specification, so nothing here is strictly wrong. But the order is load-bearing for real clients:

- The ORCA app drives a **single** radar and takes the first entry it is offered. On a dual-range radar it showed range A or range B depending on nothing the operator could see or control.
- A client that caches the list at startup locks in whichever order it happened to receive, so the choice is effectively made by a coin flip once per client process, then persists.

This surfaced while fixing the ORCA-app radar path in signalk-beluga-core, which now sorts client-side as a workaround. Sorting at the source would fix it for every consumer, and costs nothing.

## Suggested fix

Serialise the radar map from an ordered map (`BTreeMap`) or sort the keys on the way out, so `GET /radars` is stable across requests and across restarts. Radar ids are stable and sort naturally.

Worth a look at whether any other map-valued response has the same issue — `/controls` is already fine.

keesverruijt:
A client that takes the first radar from `GET /radars` got a different one from request to request. On a dual-range radar that meant range A or range B at random, with nothing the operator could see or influence — and a client that caches the list at startup locked that coin flip in for the life of its process. The ORCA app hit exactly this.

closes #497

## Cause

The response serialised its radars from a `HashMap`, whose iteration order is deliberately randomised. It now comes from an ordered map, so the order is stable across requests and across restarts, and radar ids sort naturally.

The controls endpoint was never affected, and it is worth recording why: it builds a `serde_json::Map`, which without the `preserve_order` feature is a sorted map. That is also why no other map-valued response has the problem — this one was the exception because it serialises a Rust `HashMap` through derive rather than building a JSON map.

## Testing

A unit test builds the response with two ids inserted in reverse and asserts the serialised order, so it fails deterministically without needing two radars on the network. The integration test additionally asserts that six consecutive requests agree, mirroring how the issue was observed.

## Summary

- Use `BTreeMap` to serialize `GET /radars` entries in stable natural order.
- Add unit and integration tests that verify consistent ordering across responses and requests.
- Leave the `/controls` endpoint unchanged.

### DIFF
--- a/src/bin/mayara-server/web/signalk/v2.rs
+++ b/src/bin/mayara-server/web/signalk/v2.rs
@@ -11,7 +11,7 @@ use hyper;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     net::Ipv4Addr,
 };
 use strum::EnumCount;
@@ -232,7 +232,12 @@ struct RadarsResponse {
     #[schema(example = "3.4.0")]
     version: String,
     /// Discovered radars, keyed by radar ID.
-    radars: HashMap<String, RadarApiV3>,
+    ///
+    /// Ordered, not a hash map: JSON objects carry no order of their own,
+    /// so clients that take "the first radar" would otherwise get a
+    /// different one from request to request. On a dual-range radar that
+    /// means range A or range B at random.
+    radars: BTreeMap<String, RadarApiV3>,
 }
 
 #[utoipa::path(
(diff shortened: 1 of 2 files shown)


keesverruijt:
@coderabbitai review

PR #513, opened 2026-08-09
fix(gui): stop AIS overlay subscribing to own ship

dirkwa:
## Summary

On a boat with a busy NMEA 2000 bus, the radar picture in the GUI stutters, feels sluggish, and the spokes stop filling the screen — while the radar itself is transmitting normally. Switching the renderer (WebGL ↔ WebGPU) makes no difference, because rendering was never the problem.

The AIS overlay subscribes to `vessels.*` with a wildcard path list. That matches **own ship** as well as AIS targets, and on a real boat own ship is by far the loudest context: engine, wind, water temperature and attitude all arrive at native update rate. The browser JSON-parses all of it on the main thread, which starves spoke rendering.

Measured against a live Signal K server with 128 AIS targets visible:

| | rate |
| --- | --- |
| Total deltas received by the viewer | 413/s |
| Of which own ship | 71% |
| Actual AIS targets (what the overlay draws) | ~110/s |

`handleAisDelta` already discards self and own-ship identifiers, so this data was being parsed and then thrown away.

## Approach

Send an `unsubscribe` for `vessels.self` right after the wildcard subscription, so the server stops sending own-ship deltas instead of the client discarding them.

The wildcard path list is deliberately left intact — vessel names ride the empty path, so narrowing by leaf path would silently lose them. Heading and position continue to come from `subscribeToHeading()` on `vessels.self`, which is unaffected.

Throttling with `period` was tried first and rejected: it only halved the rate (413/s → 196/s at 2000 ms, still 150/s at 30000 ms), because the volume comes from many distinct high-rate paths rather than one fast one.

## Tested

- `cargo test` — 405 passed, 0 failed
- Wire measurement against the live Signal K server: 413/s → 110/s with the unsubscribe (73% reduction)
- Rebuilt the binary and A/B'd the old and fixed GUI bundles in a real headless browser: deltas 339 → 83, spoke frames unchanged (78 vs 79), confirming the fix does not affect spoke delivery

The AIS overlay unsubscribes from `vessels.self` after subscribing to `vessels.*`. It continues to receive vessel names through the wildcard subscription. It receives own-ship heading and position through `subscribeToHeading()`. This reduces unnecessary browser processing and GUI stuttering.

### DIFF
--- a/web/gui/viewer.js
+++ b/web/gui/viewer.js
@@ -567,6 +567,24 @@ async function subscribeToAisViaSignalK() {
       subscribe: [{ path: "*" }],
     };
     socket.send(JSON.stringify(subscription));
+
+    // `vessels.*` matches own ship too, and own ship is the single loudest
+    // context on a real boat — every N2K-derived path (engine, wind, water
+    // temperature, attitude) at native update rate. Measured against a live
+    // Signal K server: 413 deltas/s total, of which 71% were own ship and
+    // only ~110/s were the AIS targets this overlay actually draws. That
+    // volume is parsed on the main thread, which starves spoke rendering —
+    // the radar picture stutters and stalls while the socket is busy.
+    //
+    // The overlay never needs own ship from here: heading and position come
+    // from subscribeToHeading() on `vessels.self`. Unsubscribing keeps the
+    // wildcard path list intact (vessel names ride the empty path, so they
+    // must not be narrowed away) while dropping the bulk of the traffic.
+    const unsubscribeSelf = {
+      context: "vessels.self",
+      unsubscribe: [{ path: "*" }],
+    };
+    socket.send(JSON.stringify(unsubscribeSelf));
   };
 
   socket.onmessage = (event) => {


dirkwa:
Follow-up verification, and one correction to the numbers above.

**Heading and position are unaffected.** They come from a separate socket with its own subscription state, so the `vessels.self` unsubscribe on the AIS socket does not touch them. Measured with both sockets open in the same order the viewer opens them, over 30 s:

- `navigation.headingTrue` — 141 deltas
- `navigation.position` — 30 deltas (matching the 1000 ms period requested)
- `navigation.headingMagnetic` — 0, because this vessel does not publish it; unrelated to this change

**Correction:** the unsubscribe does not fully suppress own-ship deltas. In the same run the AIS socket still received 396 own-ship messages alongside 1753 target messages. The overall reduction is real and reproducible (413/s → 110/s), but it is a large reduction rather than complete elimination, and the mechanism behind the residual is not yet established.

This does not change the fix's value — `handleAisDelta` discards whatever still arrives, exactly as before — but the description above reads as a cleaner cut than what is actually measured.

dirkwa:
wrong lead, sorry for the noise.

dirkwa:
Closing note — the premise did not survive measurement.

The review comment was correct: the `{context, unsubscribe}` frame is silently dropped by mayara's untagged `StreamRequest` enum, so this change was a no-op against standalone mayara. That also explains the residual own-ship deltas reported earlier.

The suggested replacement does not work either. Vessel desubscription matches by exact string against the stored pattern (`stream.rs`), and the viewer subscribes with `"*"`. So `desubscribe: [{path: "vessels.self"}]` matches nothing, and the only string that would match is `"*"` — which would remove all AIS. There is currently no way to express "all vessels except self" on this stream.

More importantly, the motivation was wrong. Profiling the real GUI against a live Signal K server with the full firehose running (298–364 deltas/s sustained over 4 minutes):

| | deltas | main thread busy | spokes |
| --- | --- | --- | --- |
| direct to mayara | 7/s | 7.3% | 1.9/s |
| via Signal K | 364/s | 14.6% | 1.9/s |

The main thread is ~85% idle in both cases, and no JavaScript function exceeds 0.3% of self time — `handleAisDelta` and `JSON.parse` never appear. Parsing 4000 real frames costs 9.2 ms in total. Spoke throughput is identical on both paths.

The wildcard subscription is genuinely wasteful — 71% of that traffic is own ship, which `handleAisDelta` discards — but it is not what makes the radar picture stutter, and this PR claimed it was. Narrowing it remains worthwhile as an efficiency change, but it needs server-side support for excluding a context and should be proposed on its own merits.

PR #520, opened 2026-08-10
fix(controls): stop numeric controls reporting a value they never had

dirkwa:
## Summary

Follow-up to #479, taking the fix to the root rather than special-casing it in the GUI as discussed there.

A control whose range the radar has never reported showed a value anyway. On a Furuno dual-range antenna, Range B has no range until its first command activates the channel, but mayara published range `0` — timestamped, and so indistinguishable from a reading. A client either drew that as a real setting or, once #515's snap-to-nearest runs over reported values, rounded it up to the shortest settable range: a range the operator never chose, on a channel that is not transmitting.

The cause is not Furuno-specific. `new_numeric` and `new_auto` passed `min_value` as the control's `default_value`, and `Control::new` seeds the value from it **and stamps it with the current time**. Every numeric control on every brand was therefore born holding its own minimum, looking observed.

Both now pass no default, so the value stays absent until something sets it. The minimum remains as `min_value`, where it belongs — a bound, not a reading.

Enum controls still seed `0` and are deliberately untouched: for an enum `0` is usually a real option (the first list entry), so dropping it is a per-control judgement rather than the clear-cut case this makes. Worth a separate look — `Power` listing `[1, 2]` is the example from #515 of why it needs care.

The second commit fixes a **pre-existing** GUI bug that this change makes reachable: a control with no value reached `innerHTML` as `undefined`, rendering the literal text "undefined" (or "undefined m" with units). It is in this PR rather than a separate one because shipping the server change alone would leave a window where absent values print as "undefined" in the panel. Verified present on `main` today, independently of the server change.

## Tested

- `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, `cargo test` — 414 passed, 0 failed
- Replay fixtures (`replay_furuno_drs2d`, `replay_navico_4g`) pass, so real captured traffic is unaffected
- Against a live DRS4D-NXT, before and after:

  | | Range B `range` | Range A `range` |
  | --- | --- | --- |
  | before | `{"value": 0}`, stamped at startup | 463 m, live |
  | after | absent | 463 m, live |

- Headless browser on the Range B panel: `main` renders "undefined", this branch renders a dash, no page errors either way

- Leave numeric and auto control values and timestamps unset until the radar reports them. Keep minimum values as validation bounds. Keep enum controls seeded to `0`.
- Display a dash for unset GUI values instead of `undefined`. Continue applying auto toggles, enabled state, allowed state, errors, callbacks, and automatic slider bounds.
- Add tests for unset numeric and auto values and timestamps.

### DIFF
--- a/src/lib/radar/settings.rs
+++ b/src/lib/radar/settings.rs
@@ -3836,6 +3836,7 @@ mod test {
         let (_, gain) = new_auto(ControlId::Gain, 0., 100., HAS_AUTO_NOT_ADJUSTABLE).take();
         assert_eq!(gain.value, None, "auto-capable numerics too");
         assert_eq!(gain.timestamp, None);
+        assert_eq!(gain.item.min_value, Some(0.), "its bound survives too");
 
         // Enum controls still seed 0 and are deliberately left alone here: for
         // an enum, 0 is usually a real option (the first list entry), so
--- a/src/lib/radar/settings.rs
+++ b/src/lib/radar/settings.rs
@@ -3824,6 +3824,10 @@ mod test {
         let (_, range) = new_numeric(ControlId::Range, 0., 100_000.).take();
         assert_eq!(range.value, None, "range must not report a phantom 0");
         assert_eq!(
+            range.timestamp, None,
+            "and no timestamp — a stamped value is what made it look observed"
+        );
+        assert_eq!(
             range.item.min_value,
             Some(0.),
             "the minimum is still a bound"
@@ -3831,6 +3835,7 @@ mod test {
 
         let (_, gain) = new_auto(ControlId::Gain, 0., 100., HAS_AUTO_NOT_ADJUSTABLE).take();
         assert_eq!(gain.value, None, "auto-capable numerics too");
+        assert_eq!(gain.timestamp, None);
 
         // Enum controls still seed 0 and are deliberately left alone here: for
         // an enum, 0 is usually a real option (the first list entry), so
--- a/src/lib/radar/settings.rs
+++ b/src/lib/radar/settings.rs
@@ -2455,7 +2455,11 @@ pub(crate) fn new_numeric(control_id: ControlId, min_value: f64, max_value: f64)
     let control = Control::new(ControlDefinition::new(
         control_id,
         ControlDataType::Number,
-        min_value,
+        // No default: the minimum is a bound, not a reading. Seeding it made
+        // the control indistinguishable from one the radar had actually
+        // reported — a Furuno dual-range B channel published range 0 with a
+        // timestamp before it was ever activated.
+        None,
         None,
         false,
         min_value,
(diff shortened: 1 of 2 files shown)


dirkwa:
Fixed a real bug in my own GUI change, found by a local `cr` run — the CodeRabbit check on this PR is green but has not reviewed it at all (no reviews, no inline comments), so the tick means "did not run" rather than "looks fine".

The unset branch `return`ed early, which skipped everything after the value rendering: the auto toggle, the enabled checkbox, the allowed/disallowed disabling, error display, and the control callbacks — whose comment explicitly says "Always notify". None of those depend on a value having arrived. On a Furuno Range B, where gain/sea/rain are auto-capable, the auto toggles lost their state entirely.

Now only the value rendering is skipped and the rest still runs.

Verified in a headless browser against a live DRS4D-NXT:

| | Range B | Range A |
| --- | --- | --- |
| "undefined" in panel | no | no |
| auto toggles rendered | 4 | 4 |
| dashes for unset values | 7 | 4 |
| page errors | none | none |

`cargo fmt --check` clean, `cargo test` 414 passed / 0 failed.

dirkwa:
@coderabbitai review

dirkwa:
Both addressed.

**Auto-adjustable bounds** — valid, and there was a second problem alongside it. The `i.min`/`i.max` assignments sat inside the description branch, so besides being skipped when `unset`, they were also gated on a *description element* existing, which has nothing to do with the bounds. Moved out of that branch entirely: bounds follow the auto mode, not the value. Only value and description rendering stay conditional on `unset`. The `"A+n"` description path is unchanged — still only when `descriptions[value]` misses and auto is on.

**`gain.item.min_value`** — added.

Verified in a headless browser against a live DRS4D-NXT, both channels:

| | Range B (never reported) | Range A (live) |
| --- | --- | --- |
| sliders with min/max | 22 / 22 | 22 / 22 |
| auto toggles | 4 | 4 |
| "undefined" in panel | no | no |
| page errors | none | none |

`cargo fmt --check` clean, `cargo clippy --all-targets -D warnings` clean, `cargo test` 414 passed / 0 failed.

dirkwa:
@coderabbitai review

dirkwa:
@coderabbitai review

dirkwa:
Declining this one, with a reason.

The literals it flags are the existing convention in this file, not something this PR introduces. `new_numeric(ControlId::Range, 0., 100_000.)` and `new_auto(ControlId::Gain, 0., 100., …)` both already appear on `main` in `reported_auto_defaults_false_for_auto_capable_controls`, directly above the test added here — my test was written to match its neighbour. Production code does the same (`new_numeric(ControlId::Spokes, 0., 9999.)`).

In these tests the upper bound is throwaway fixture scaffolding: nothing asserts against it, and the number carries no meaning beyond "some upper bound". Naming it would add indirection without adding information, and would leave the adjacent pre-existing test inconsistent with it.

The "no magic numbers" guideline is aimed at values whose meaning is not obvious at the point of use. If the maintainers would rather see these named, it is worth doing across the file in its own pass rather than only on the two lines this PR happens to touch — happy to open that separately.

PR #530, opened 2026-08-11
refactor(api): drop the dual and dualGroup radar fields

keesverruijt:
No user-visible change: the radar list simply stops carrying two fields that nothing consumes.

`dual` and `dualGroup` were added so a client could tell which two radar entries are the ranges of one antenna. They are not needed for that. The radar list is serialised in a stable order that keeps those entries adjacent, so a client pairs them by position — the GUI does exactly that as of #529 and reads neither field.

Removing them puts the radar list back on the surface shared with signalk-server, leaving `replay` as the only documented mayara extension in the API-parity allowlist.

## Implementation

- `RadarApiV3` loses both fields, along with `RadarInfo::base_key()` and its private helper, which existed only to compute `dualGroup`.
- `RadarInfo::dual` itself stays — it is internal state the brands rely on, and the 4G replay test still asserts it.

## Test plan

- [x] `cargo test` passes, including the API-parity test against the signalk-server checkout.
- [x] `cargo test --features pcap-replay --test replay_navico_4g` passes.
- [x] `cargo fmt` and `cargo clippy --no-deps --all-targets -- -D warnings` clean.

> Stacked on #529 (based on that branch, so merge it first). Rebasing onto `main` after #529 lands leaves the diff unchanged.


keesverruijt:
@coderabbitai review

PR #533, opened 2026-08-12
fix(gui): send the auto adjustment as a number

### LINKED ISSUE #466: HALO 20+ Sea auto/manual changes are not reflected consistently in stream
### Mayara version or commit ID

681d8ac462c360e2b6479369b5a8ec08d8f021b3

### Radar brand

Navico

### Radar model

HALO20+

### PCAP file

_No response_

### Network diagnostics

_No response_

### Screenshot or video

_No response_

### Describe the issue

On Navico HALO 20+, Sea state updates are not streamed symmetrically. When Sea changes from manual to auto, the update appears in the stream. But when Sea changes back from auto to manual, the stream does not always update, or updates incorrectly. As a result, the streamed control state can drift away from the radar’s actual state.

**Steps to reproduce:**
1. Connect a Navico HALO 20+.
2. Switch Sea from manual to auto.
3. Verify that the stream updates.
4. Switch Sea back from auto to manual.
5. Check whether the stream updates again.
6. Expected behavior:
7. Any change to Sea should be reflected in the stream in both directions: manual -> auto and auto -> manual.

**Actual behavior:**
The manual -> auto transition updates the stream, but the reverse transition does not always do so.

keesverruijt:
Dragging the adjustment slider of a control that is on auto did nothing whatsoever. On a HALO that is Sea clutter's -50..+50 adjustment: the slider moved, the radar carried on unchanged, and the control fell back to showing as unconfirmed. Nothing reached the radar, because nothing reached mayara — the browser's request was rejected before any handler saw it.

Verified on a HALO24: the adjustment now takes effect, and the radar reports it back.

## Root cause

A range input hands out its position as a string. `do_change` converted that to a number only for controls that carry `user_units`, and Sea clutter has none, so the string went out untouched.

The server takes `value` as any JSON and coerces it, which is why the manual slider worked with the same string. `autoValue` is an `Option<f64>` and is strict, so the auto adjustment 422'd on every drag:

```
PUT …/controls/sea  {"auto":true,"autoValue":"-26"}
→ 422  autoValue: invalid type: string "-26", expected f64
```

An auto adjustment is numeric by definition, so it is now converted at the point it is put into the request. `value` is left alone — string-valued controls go through the same path, and the server already handles the numeric ones.

## Related

Separate from #532, which fixes the byte encoding of the same command. That one was masked by this: the request never arrived, so the encoding was never exercised from the GUI. Both are needed for the adjustment to work, and both are confirmed on the water. Neither closes #466, which is about mode changes reaching the stream.

### DIFF
--- a/web/gui/control.js
+++ b/web/gui/control.js
@@ -1587,8 +1587,12 @@ function do_change(v) {
   update.auto = auto;
   message.auto = auto;
   if (auto && control.hasAutoAdjustable) {
-    update.autoValue = storeValue;
-    message.autoValue = value;
+    // An auto adjustment is always a number, and the server only accepts one
+    // there — unlike `value`, which it takes as any JSON. A slider hands us
+    // its position as a string, so a control without units (Sea clutter, for
+    // one) would otherwise send `"-26"` and have the whole request rejected.
+    update.autoValue = Number(storeValue);
+    message.autoValue = Number(value);
   } else {
     update.value = storeValue;
     message.value = value;


keesverruijt:
Superseded by #534, which fixes this on the server side instead.

The GUI is embedded in the server binary, so a browser always talks to the server that served it — there is no version skew for a client-side coercion to protect against. Keeping this would also leave the GUI half-coerced: `value` still goes out as a slider string, and the only reason to treat `autoValue` differently was the strictness that #534 removes.

PR #570, opened 2026-08-22
docs(raymarine): explain which addresses a Raymarine network uses

keesverruijt:
The Raymarine help page told users that any Raymarine radar needs DHCP and has no fallback address. That is true of a Quantum, but wrong for an RD or HD radome — those give themselves a fixed `10.x.x.x` address, ignore DHCP entirely, and keep that address forever. It is the reason an RD radar paints a perfect picture while every control does nothing, so the page was steering affected users away from their actual problem.

Also adds the address ranges a Raymarine network uses, which were not documented anywhere. The network must be `198.18.0.0/21`; chartplotters claim `198.18.0.32`–`198.18.3.255` for themselves without checking whether anything already has it; `198.18.0.2`–`198.18.0.19` is where a hand-configured machine belongs; and `198.18.4.0`–`198.18.7.254` is what a third-party DHCP server may hand out. We have a user capture with the Mayara host sitting in the chartplotter range, which is a collision waiting to happen.

Ranges are per Raymarine's own [NETWORKING - IP / Netmask / DHCP](https://support.raymarine.com/s/article/NETWORKING---IP---Netmask---DHCP?language=en_US) FAQ, and every `198.18` address across our captures — 23 of 23 — falls inside the documented `/21`.

Separately, drops a stale line claiming `232.1.1.1:5800` is only joined when `--allow-wifi` is given; both beacon groups have been joined unconditionally since #563.

Docs only — no code changes.

### DIFF
--- a/docs/raymarine-setup.md
+++ b/docs/raymarine-setup.md
@@ -8,7 +8,7 @@ The radar and the Mayara machine must be on the same wired Ethernet network.
 
 A **Quantum** needs a DHCP server on that network: it has no fallback address and stays silent until it gets a lease. An **RD or HD radome** never asks for one — it gives itself a fixed `10.x.x.x` address and keeps it forever, which is why it needs the Mayara machine to hold an address in that range too.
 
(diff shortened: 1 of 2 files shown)


keesverruijt:
@coderabbitai review

PR #574, opened 2026-08-22
fix(network): pin multicast sends to the intended interface

keesverruijt:
On a boat where the Mayara machine has more than one network — WiFi and Ethernet, or a cellular dongle holding the default route — Mayara could fail to find radars on the radar network, and fail to control the ones it did find. The symptom looks like a flaky cable or a switch problem: a radar that appears sometimes, or appears but ignores every command.

Multicast datagrams were leaving by whichever interface held the route, rather than the interface meant for them, because the send socket never set `IP_MULTICAST_IF`.

Worst affected is **discovery**, for every brand. The locator builds one socket per interface address and sends a beacon request through each, intending to probe them all — but all of them left by the same interface, so only one network was ever really probed.

Per brand:

- **Navico** and **Garmin** command their radars over multicast, so commands were lost the same way.
- **Raymarine** commands are unicast and were unaffected, but its beacons, wake burst and NavData are multicast — so a radar could go undiscovered, never wake, or never receive heading and position.

The fix is one `set_multicast_if_v4` call on the send socket. It needs no privileges and is available on every target we build, Windows included. Unicast is unaffected by the option — for unicast the outgoing interface still follows the routing table, which is what #569 is about.

Also renames `create_multicast_send` to `create_connected_send`. It does nothing multicast-specific — it binds to the given NIC address and connects — and half its callers pass a unicast radar address. The old name helped hide this bug, and made the unicast callers look like they were doing something they were not.

## Summary

Pins multicast UDP sends to the selected network interface with `IP_MULTICAST_IF`.

Renames `create_multicast_send` to `create_connected_send` for multicast and unicast destinations.

Updates Garmin, Navico, Raymarine, and locator senders to use the renamed helper.

Updates documentation and log messages to describe connected-send behavior.

Adds tests for multicast interface selection and reliable local-port allocation.

### DIFF
--- a/src/lib/brand/garmin/cdm_heartbeat.rs
+++ b/src/lib/brand/garmin/cdm_heartbeat.rs
@@ -163,7 +163,7 @@ pub(crate) async fn run(syc_group_id: Arc<AtomicU8>) {
 }
 
 async fn send_one(packet: &[u8], dest: &SocketAddrV4, nic: &Ipv4Addr) {
-    match network::create_multicast_send(dest, nic) {
+    match network::create_connected_send(dest, nic) {
         Ok(sock) => {
             // Suppress local loopback: otherwise the kernel delivers
             // our own packet right back to mayara's receive socket on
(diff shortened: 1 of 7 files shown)


keesverruijt:
@coderabbitai review

keesverruijt:
@coderabbitai review

PR #617, opened 2026-08-31
fix(windows): detect wireless and unusable interfaces correctly

keesverruijt:
On Windows, mayara searched for radars over WiFi even when `--allow-wifi` was not given, so a laptop on a marina hotspot would flood its wireless link with discovery traffic that the flag was meant to suppress. It also probed adapters that can never carry radar data at all — Bluetooth personal area networks, VPN tunnels (Teredo, 6to4, IP-HTTPS), and dial-up/PPP links.

Now WiFi is skipped unless `--allow-wifi` is given, and links that cannot carry a radar's spoke stream are skipped in either mode.

## Technical detail

The WLAN check compared an adapter's *friendly* name, which is what `NetworkInterface::show()` reports (`WiFi`), against `WLAN_INTERFACE_INFO::strInterfaceDescription` (`Broadcom 802.11ac Network Adapter`), so it never matched. The NUL padding on that fixed-size field was not stripped either — `str::trim` does not remove `\0` — so the comparison could not have succeeded even for equal names. `is_wireless_interface` therefore always returned `false`.

Classification now goes through `GetIfTable2`. `MIB_IF_ROW2` carries the friendly name (`Alias`), the adapter type and `PhysicalMediumType` together, and all three are needed: Windows reports a Bluetooth PAN as `IF_TYPE_ETHERNET_CSMACD`, identical to real Ethernet, so an `IfType` allowlist alone does not exclude it. The shared `LinkKind` enum replaces the `is_wireless_interface` predicate on all three platforms; Linux and macOS keep their existing WiFi probe and never report `Unusable`.

Skipped interfaces are reported as `LinkTypeIgnored` in the diagnostics API, and an interface missing from the table falls back to `Wired`, so a lookup failure can never silently hide a radar.

Two latent faults go away with the old code: it panicked on any WLAN API error other than a stopped service, and it indexed `WLAN_INTERFACE_INFO_LIST::InterfaceInfo` — a one-element flexible array head — out of bounds on a machine with two or more wireless adapters.

## Test plan

- [x] Six unit tests cover the classification table and the NUL-padding decode
- [x] `cargo test` passes (480 lib + 16 integration)
- [x] Verified on hardware with a WiFi and a Bluetooth PAN adapter: default run skips both, `--allow-wifi` searches WiFi and still skips Bluetooth
- [ ] Not built on Linux or macOS — those two files only wrap their existing `bool` in `LinkKind`

> [!NOTE]
> Stacked on #616, without which the pre-commit hook cannot pass on Windows. Review only the top commit; the base retargets to `main` once #616 merges.

https://claude.ai/code/session_01H85CDR2cLqoPEm2UpQazcA

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -114,7 +114,7 @@ rtnetlink = "0.23.0"
 
 [target.'cfg(target_os = "windows")'.dependencies]
 w32-error = "1.0.0"
-windows = { version = "0.62.2", features = ["Win32_NetworkManagement_IpHelper", "Win32_NetworkManagement_WiFi", "Win32_Networking_WinSock", "Win32_System", "Win32_System_Threading", "Win32_Security", "Win32_System_Diagnostics", "Win32_System_Diagnostics_Debug", "Win32_System_IO", "Win32_NetworkManagement_Ndis", "Win32_Networking_WinSock", "Win32_System_SystemServices" ] }
+windows = { version = "0.62.2", features = ["Win32_NetworkManagement_IpHelper", "Win32_Networking_WinSock", "Win32_System", "Win32_System_Threading", "Win32_Security", "Win32_System_Diagnostics", "Win32_System_Diagnostics_Debug", "Win32_System_IO", "Win32_NetworkManagement_Ndis", "Win32_Networking_WinSock", "Win32_System_SystemServices" ] }
 
 [build-dependencies]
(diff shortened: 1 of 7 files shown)


keesverruijt:
@coderabbitai review

keesverruijt:
@coderabbitai review

PR #618, opened 2026-08-31
fix(windows): keep the OVERLAPPED alive and re-arm address changes

keesverruijt:
On Windows, mayara stops noticing network changes after the first one. Plug in the radar's Ethernet cable, or let DHCP renew to a new address, and the server keeps listening on the old interface set instead of picking up the new one — and it does so while spinning a CPU core, because the same stale notification is processed over and over.

## Technical detail

Two faults in the same registration:

**The OVERLAPPED did not outlive the request.** `NotifyAddrChange` was armed against a structure on the stack of `create_ip_addr_change_event`, which returned immediately afterwards. The I/O manager writes the completion status into that structure when an address changes, so the write landed in a dead stack frame. Nothing cancelled the pending request either, so it stayed armed against that memory for the life of the process.

**The registration was never renewed.** `NotifyAddrChange` fires once per registration. The event is manual-reset and was never reset, so after the first address change it stayed signaled permanently, `WaitForMultipleObjects` returned `WAIT_OBJECT_0` on every iteration, and the loop broadcast a change notification as fast as it could schedule — driving a full interface re-scan each time.

The `OVERLAPPED` now lives in the wait loop that owns it, arming is a separate step repeated after every notification (arming also resets the event, as the request completes asynchronously), and `CancelIPChangeNotify` tears the pending request down before the structure goes out of scope. The pointer handed to the kernel keeps the provenance of a mutable borrow, since the kernel writes through it even though the binding is `*const`. A `WaitForMultipleObjects` failure now breaks out of the loop rather than spinning on the same error forever.

## Test plan

- [x] New test covers the cancellation path end to end: the watcher arms, is cancelled, and returns
- [x] `cargo test` passes (481 lib + 16 integration)
- [x] Server starts, arms the notification, and reports no spurious change events while idle

The re-arm path itself is not exercised by a test: triggering a real address change needs an adapter reconfiguration I could not perform here, so that path was reviewed by inspection against the `NotifyAddrChange` contract.

https://claude.ai/code/session_01H85CDR2cLqoPEm2UpQazcA

- Keeps the Windows `OVERLAPPED` structure alive while address-change notifications are pending.
- Re-arms `NotifyAddrChange` after each notification and resets the event before re-arming.
- Cancels pending notifications before cleanup.
- Exits the wait loop when a wait or registration fails.
- Adds cancellation coverage for watcher shutdown.
- Passes formatting, clippy, startup, and idle behavior checks.

### DIFF
--- a/src/lib/network/windows/mod.rs
+++ b/src/lib/network/windows/mod.rs
@@ -5,15 +5,17 @@ use tokio::sync::broadcast;
 use tokio_util::sync::CancellationToken;
 use w32_error::W32Error;
 use windows::Win32::Foundation::{
-    CloseHandle, ERROR_IO_PENDING, ERROR_SUCCESS, HANDLE, WAIT_EVENT,
+    CloseHandle, ERROR_IO_PENDING, ERROR_SUCCESS, HANDLE, WAIT_OBJECT_0,
 };
-use windows::Win32::NetworkManagement::IpHelper::NotifyAddrChange;
+use windows::Win32::NetworkManagement::IpHelper::{CancelIPChangeNotify, NotifyAddrChange};
 use windows::Win32::NetworkManagement::Ndis::{
     NDIS_PHYSICAL_MEDIUM, NdisPhysicalMediumBluetooth, NdisPhysicalMediumNative802_11,
     NdisPhysicalMediumWirelessLan, NdisPhysicalMediumWirelessWan,
 };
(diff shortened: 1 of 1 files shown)


keesverruijt:
@coderabbitai review

keesverruijt:
@coderabbitai review

keesverruijt:
@coderabbitai full review

keesverruijt:
@coderabbitai full review

PR #648, opened 2026-09-04
fix(stream): honour subscription terms on a self or all stream

### LINKED ISSUE #647: fix(stream): subscription terms are ignored on a self or all stream
## Symptom

A client connected with the default `subscribe=self` (or `subscribe=all`) cannot pace anything it receives. It can ask — `policy`, `period` and `minPeriod` are accepted without complaint — but every value still arrives as fast as the boat produces it. A chart app on a metered or high-latency link has no way to say "once a second is enough" unless it opens the stream with `subscribe=none` and then names every path it wants, which is exactly the work the baseline exists to save.

This applies to radar controls as well as navigation, so it is not fixed by #634 / #646, which pace both — but only on a `subscribe=none` stream.

The embedded GUI is unaffected: it opens every stream with `subscribe=none`.

## Why

The mode is checked before the terms are, in every one of the three places that decide delivery (`src/lib/stream.rs`):

```rust
// is_subscribed / is_subscribed_path
match self.mode {
    Subscribe::All => return true,
    Subscribe::SelfOnly => return true,   // (unless the path is `vessels.`)
    Subscribe::None => {}
}

// is_due_periodic
if self.mode != Subscribe::None {
    return false;
}
```

So a `self` client short-circuits to "yes, on change" and never reaches `terms_for` / `allow_now`. The baseline is being treated as an answer rather than as the default for a path the client said nothing about.

`send_all_subscribed` (`src/bin/mayara-server/web/signalk/v2.rs`) has the same shape — it filters the control values it gathers only when `mode == Subscribe::None`.

## Approach

An explicit request should replace the baseline for the paths it covers, rather than being ignored underneath it: resolve terms first, and fall back to the baseline only when no request names the path. Then `self` + a `fixed` subscription means periodic delivery, `self` + `minPeriod` means a rate limit, and `self` with nothing named behaves exactly as it does today.

This changes delivery for controls as well as navigation, which is why it is separate from #646.

## Not in scope

AIS pacing. A `vessels.*` delta is subscribed as a whole by context (`is_subscribed_ais`), not by leaf path, so pacing it needs a different mechanism than the per-path throttle and should be judged on its own.

keesverruijt:
A client on the default `subscribe=self` can now ask for data at the rate it wants and actually get it. Before this, `policy`, `period` and `minPeriod` were accepted on such a stream and then ignored, so a chart app on a metered or high-latency link had only one way to slow anything down: reopen the stream as `subscribe=none` and name every single path it needed — exactly the work the baseline exists to save.

This covers radar controls as well as navigation. #646 paced both, but only on a `subscribe=none` stream.

A client that names no subscriptions sees no change at all, and the embedded GUI is unaffected — it opens every stream with `subscribe=none`.

## Cause

The `?subscribe=` mode was checked before the terms were, in each of the three places that decide delivery, so a `self` or `all` client short-circuited to "yes, on change" and never reached `terms_for` / `allow_now`. The baseline was answering for paths the client had explicitly spoken about.

## Approach

Terms resolve first; the baseline becomes the fallback for a path no request names (`baseline_carries`). An explicit subscription replaces the baseline for the paths it covers instead of being ignored underneath it. With the terms lookup now common to all paths, the separate control / navigation / event branches in `is_subscribed_path` collapse into a single `is_paced_path` check. `send_all_subscribed` drops its `mode == Subscribe::None` guard and always filters, which is now correct in every mode.

## Behaviour change

An existing `subscribe=self` client that sends a subscription with a policy was getting a firehose and will now get the rate it asked for. That is the fix, but it is a change in what such a client receives.

One test from #646 asserted the old behaviour (`a_self_client_is_never_served_from_the_schedule`) — that assertion is the bug here, so it is replaced. Three tests are added: a `self` client naming nothing is served everything own-ship on change and never from the schedule, terms pace only the paths they cover, and an `all` client can pace controls while AIS still rides its baseline.

## Not in scope

AIS pacing. A `vessels.*` delta is subscribed as a whole by context, not by leaf path, so it needs a different mechanism than the per-path throttle.

closes #647

Enables `policy`, `period`, and `minPeriod` terms for `subscribe=self` and `subscribe=all` streams.

- Applies explicit terms to named navigation and radar paths.
- Uses the `subscribe` baseline for unnamed paths.
- Applies pacing to radar controls in `send_all_subscribed`.
- Preserves existing behavior for clients without named subscriptions.
- Keeps AIS pacing out of scope.
- Updates pacing documentation and adds coverage for self, path-specific, and all-mode subscriptions.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -254,8 +254,9 @@ are events: each carries a transition that cannot be inferred from the next (an
 alarm clearing, a target lost), so they are always delivered as they happen,
 whatever policy is asked for.
 
-These options govern a stream opened with `subscribe=none`. A `self` or `all`
-stream is served from its baseline as values change.
+These options apply on any stream. A path a subscription names is delivered on
+the terms it names; a path no subscription mentions is left to the
+`?subscribe=` baseline, which delivers as values change.
 
 ### Client → Server: Unsubscribe
 
--- a/src/bin/mayara-server/web/signalk/v2.rs
+++ b/src/bin/mayara-server/web/signalk/v2.rs
@@ -2189,11 +2189,10 @@ async fn send_all_subscribed(
     for radar in radars.get_active() {
         rcvs.append(&mut radar.controls.get_radar_control_values());
     }
-    // Under `none`, keep only explicitly-subscribed controls; `self`/`all` get
-    // them all (radar controls are own-ship data, always in those baselines).
-    if subscriptions.mode == Subscribe::None {
-        rcvs.retain(|x| subscriptions.is_subscribed(x, true));
-    }
+    // Keep what the client is owed: the controls it named, on the terms it
+    // named them on, plus everything its baseline carries (radar controls are
+    // own-ship data, so `self` and `all` carry them all).
+    rcvs.retain(|x| subscriptions.is_subscribed(x, true));
     log::debug!("Sending {} subscribed controls", rcvs.len());
     if !rcvs.is_empty() {
         let mut delta: SignalKDelta = SignalKDelta::new();
--- a/src/lib/stream.rs
+++ b/src/lib/stream.rs
@@ -604,6 +604,17 @@ fn is_navigation_path(path: &str) -> bool {
     )
 }
 
+/// Whether `path` carries a value that stands between updates, and so can be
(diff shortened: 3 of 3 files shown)


keesverruijt:
@coderabbitai review

PR #659, opened 2026-09-07
fix(replay): count listeners, not the addresses they listen on

### LINKED ISSUE #656: fix(replay): the re-send pass is skipped when a listener joins a known address
## Symptom

Replaying a capture can deliver nothing to a radar that was discovered from that same capture. The dispatcher is meant to make a second pass once discovery has registered its listeners — that is how a radar gets the reports and spokes that were replayed before it existed — but the pass is sometimes skipped, and the run ends having sent that radar nothing.

It shows up as a replay session where the radar appears in the API with its model and ranges, and no data ever follows.

## Why

`src/lib/replay.rs` counts listeners two different ways. At the top of each pass:

```rust
let listeners_before: usize = state.channels.lock().unwrap()
    .values().map(|v| v.len()).sum();          // sum of listeners
```

and then, deciding whether to make the extra pass:

```rust
let listeners_now = state.channels.lock().unwrap().len();   // number of addresses
if listeners_now > listeners_before {
    continue;
}
```

One is the total number of listeners, the other the number of multicast addresses that have any. A listener that registers on an address already in the map — which is the normal case, since discovery and the data loop share groups — raises the sum but not the key count, so the comparison fails and the pass is skipped.

The wait loop just above gets it right, computing the sum and breaking as soon as it rises. So the dispatcher correctly notices the new listener and then asks the wrong question about it.

## Scope

Pre-existing, from #116. Only `--repeat` runs hide it, because they loop the capture anyway — which is why the replay tests that assert decoded spokes (#655) set `repeat: true` and are unaffected.

Raised by CodeRabbit while reviewing #655.

keesverruijt:
Replaying a capture could deliver nothing to a radar that was discovered from that same capture: it appears in the API with its model and ranges, and no data ever follows.

The dispatcher is meant to make a second pass once discovery has registered its listeners — that is how a radar receives the reports and spokes replayed before it existed — and the pass was sometimes skipped.

## Cause

The decision compared two different things. At the start of a pass the dispatcher counted listeners:

```rust
let listeners_before = ...values().map(|v| v.len()).sum();
```

and then decided whether to repeat by counting *addresses*:

```rust
let listeners_now = state.channels.lock().unwrap().len();
if listeners_now > listeners_before { continue; }
```

A listener that joins a group another one already holds raises the first and not the second. That is the ordinary case, because discovery hands a radar the groups the locator is already listening on. The wait loop immediately above computed the sum correctly and broke as soon as it rose — so the dispatcher noticed the new listener and then asked the wrong question about it.

## Approach

One `listener_count()` on `ReplayState`, used in all four places that were counting by hand, so the two readings cannot drift apart again.

## Tests

Three, over a state built with listeners spread across addresses — including several sharing one, which is the case that was miscounted. Checked against the old expression: the two that matter fail.

Pre-existing, from #116. `--repeat` runs were unaffected, since they loop the capture anyway.

closes #656

Fixes replay dispatch so a second pass runs when discovery adds a listener to an existing address.

Adds `ReplayState::listener_count()` and uses it to count listeners across all addresses. Adds unit and end-to-end tests for listeners sharing addresses and receiving replayed data.

### DIFF
--- a/src/lib/replay.rs
+++ b/src/lib/replay.rs
@@ -112,6 +112,23 @@ struct ReplayState {
     channels: Mutex<HashMap<SocketAddrV4, Vec<mpsc::Sender<ReplayPacket>>>>,
 }
 
+#[cfg(feature = "pcap-replay")]
+impl ReplayState {
+    /// How many listeners are registered, across every address.
+    ///
+    /// Counting addresses instead — `channels.len()` — misses a listener that
+    /// joins a group another one already holds, and that is the ordinary case:
+    /// discovery hands a radar the groups the locator is already listening on.
+    fn listener_count(&self) -> usize {
+        self.channels
+            .lock()
+            .unwrap()
+            .values()
+            .map(|v| v.len())
+            .sum()
+    }
+}
+
 /// Initialize the replay system with a pcap/nnd file. Called once at startup.
 #[cfg(feature = "pcap-replay")]
 pub fn init(path: &Path) -> io::Result<()> {
@@ -215,13 +232,7 @@ pub async fn run(realistic_timing: bool, repeat: bool, max_time: Option<u32>) {
 
     let mut first_pass = true;
     loop {
-        let listeners_before: usize = state
-            .channels
-            .lock()
-            .unwrap()
-            .values()
-            .map(|v| v.len())
-            .sum();
+        let listeners_before = state.listener_count();
         let mut prev_ts = Duration::ZERO;
         let mut sent = 0u64;
         let mut unrouted = 0u64;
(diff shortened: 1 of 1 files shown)


keesverruijt:
Added in 6b2f0dc — the end-to-end test the risk note asks for, and it was a fair ask: the unit tests covered the counting rather than the behaviour that depends on it.

I had judged this untestable when opening the PR, on the grounds that the dispatcher runs off a process-global. That was wrong. `run()` is public and `create_listen` is `pub(crate)`, so a unit test inside the crate can drive the real dispatcher, and the global being per-process is fine as long as one test owns it.

The test replays a one-packet capture with a listener already on the group — the locator — then registers a second listener on that same group after the first pass has been and gone, standing in for discovery handing a radar the group the locator already holds. It asserts the second listener still receives the capture. Run against the original `channels.len()` decision it fails with "the second pass never came: a listener joining a known group received nothing", so it pins the behaviour rather than the implementation.

keesverruijt:
@coderabbitai review

PR #670, opened 2026-09-08
feat(koden): let an unwatched radar stand down after a set time

### LINKED ISSUE #665: feat(koden): let an unwatched radar stand down
A Koden radar keeps transmitting for as long as mayara runs, even with nobody looking at it: mayara sends the radar a keepalive every 10 s regardless of whether any client is watching. The **Auto standby** setting introduced in #663 should apply to Koden radars as it does to Navico: once nobody has watched the radar for the set time, mayara stops holding it up.

## Technical notes

- Offer the control in `brand/koden/settings.rs` via `new_auto_standby()`; read `RadarInfo::stand_down()` on the existing keepalive tick in `brand/koden/report.rs` (`send_keepalive()`, `KEEPALIVE_INTERVAL_SECS`) and skip the packet while standing down, logging the edge once.
- Whether a Koden radar actually leaves Transmit when the keepalive stops is **unverified**; needs a test on hardware before the control is offered.

Part of #633.

keesverruijt:
A Koden radar kept transmitting for as long as mayara ran, even with nobody looking at it: mayara sent the radar its keep-alive every ten seconds regardless of whether any client was watching. The **Auto standby** setting from #663 (Off / 1 / 5 / 15 / 30 min, default 1 min) now applies to Koden radars too: once nobody has watched the radar for that long, mayara stops sending the keep-alive and lets the radar stand down; the moment a viewer comes back the keep-alive resumes. Another controller on the network keeps the radar up with its own keep-alive, so nothing changes for it.

Not verified on Koden hardware: the protocol notes only record that the radar acknowledges the keep-alive, not what it does when it stops. If a Koden radar turns out to keep transmitting without it, this brand needs the Furuno approach from #669 (an explicit standby for a transmit that was asked through mayara).

## Implementation

`new_auto_standby()` in `koden/settings.rs`, and the 10 s tick in `koden/report.rs` skips `send_keepalive()` while `RadarInfo::stand_down()` is set, logging the transition once each way. The tick cadence is unchanged.

Closes #665

https://claude.ai/code/session_015VvpP3eXB2WhgD8bwyDWoU

### DIFF
--- a/docs/internals/radar-status.md
+++ b/docs/internals/radar-status.md
@@ -120,7 +120,10 @@ the radar never stands down. Per brand:
   memory is set only by a Transmit request that reached the radar through mayara and
   cleared by a Standby request through mayara or by the radar reporting anything but
   Transmit.
-- **Koden, Garmin**: not yet; see #665, #667.
+- **Koden**: the radar answers a keep-alive mayara sends every 10 s; it is left out while
+  standing down. Whether the radar then leaves Transmit by itself is still to be confirmed
+  on hardware (#665).
+- **Garmin**: not yet; see #667.
 
 ## ARPA counts as a subscriber — for idle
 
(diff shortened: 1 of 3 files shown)


keesverruijt:
@coderabbitai full review

PR #672, opened 2026-09-08
refactor(furuno): use the shared transmit claim helper

keesverruijt:
No user-visible change: a Furuno radar stands down exactly as before.

#671 hoisted the "is this transmit ours to stand down" request rule into `radar/mod.rs` for Garmin, leaving Furuno's identical private copy behind. This drops that copy, names the dual-range ids with the shared `DUAL_RANGE_A`/`DUAL_RANGE_B`, and removes the unit tests that now duplicate the ones on the shared helper. Furuno's report-side release (any range leaving Transmit ends the claim) is deliberately untouched.


keesverruijt:
@coderabbitai review

PR #682, opened 2026-09-16
refactor(garmin): encode and decode packets with deku

keesverruijt:
Nothing changes in what a Garmin radar shows or does — with one exception: a malformed packet can no longer take the report receiver down.

## Why

mayara read Garmin packets by indexing datagrams at offsets kept in comments, and built every outgoing packet by writing bytes into a buffer the same way. This converts the brand to [deku](https://crates.io/crates/deku), following Navico in #680, so a wire layout is declared once instead of being spelled out in arithmetic at each site. Raymarine and Furuno follow in their own PRs.

## How

One commit per packet kind:

- **Scalar reports** — most Garmin reports carry a single setting, and it is the header's length field, not the opcode, that says whether the value is one, two or four bytes. A width we have no layout for still reads as zero, as it always has.
- **Spoke and status packets** — the enhanced spoke header, the HD spoke header and the HD state report, whose unread bytes are now visible in the declaration rather than skipped silently.
- **Discovery and capability** — heartbeat, product data, the 320-bit capability bitmap and the range table. The heartbeat still reads in two parts, so a body that stops before the identifier still says who is talking, and the range table still checks its entry count before reading the list that count sizes.
- **Outgoing packets** — the three command shapes, the two commands that bypassed those builders, the heartbeat mayara broadcasts and the two CDM frames. The opcode stays data rather than becoming a variant: the same three shapes carry dozens of settings.
- **One shared header** — both dispatchers picked the opcode and length out of the datagram by hand before deciding what a packet was, and the encoders had grown their own copy of the same two fields.

### The crash

A status packet declaring a four-byte value while carrying fewer bytes was sliced at the declared width, which panicked and took the report receiver with it — reachable from any malformed datagram on the report port. A payload shorter than the width it declares is now a decode error like any other bad packet. There is a regression test.

### Hot path

A spoke's header used to be read twice: once to peek at the range indicator that picks the antenna, then again in full. It is decoded once and passed down. Sample payloads remain borrowed slices, and HD's 1-bit unpacking stays a hand-written single pass — it maps each bit through the legend table as it goes, which a declarative read would turn into an extra allocation and an extra pass on the busiest path in the brand. (`bitvec` is already in the tree via `nmea-parser`, so that is a performance choice, not a dependency one.)

## Testing

- New unit tests pin the spoke and status offsets and the scalar widths; nothing covered those paths before.
- The tests that pin command frames and the outgoing heartbeat byte for byte are unchanged and still pass — the bytes on the wire are the same.
- Both replay fixtures pass: xHD, and Fantom Pro dual-range which covers Range B. Full `cargo test` is green.
- **Not tested against hardware.** There is no Garmin radar available here, so unlike #680 — which ran on real radars — the evidence for this one is the captured fixtures and the byte-level tests.

https://claude.ai/code/session_01Pqsda2PAjcTDUhPAzMes6J

## Summary

- Refactors Garmin packet encoding and decoding to use `deku` wire-layout structs.
- Covers headers, reports, spoke and status packets, discovery, capabilities, range tables, commands, and heartbeat frames.
- Returns decode errors for truncated payloads instead of panicking.
- Reuses decoded spoke headers and preserves optimized borrowed sample and HD bit unpacking.
- Adds tests for packet offsets, scalar widths, command layouts, and malformed packets.

Replay fixtures pass, and `cargo test` is green. Hardware testing is not performed.


keesverruijt:
@coderabbitai full review

keesverruijt:
@coderabbitai full review

keesverruijt:
@coderabbitai full review

PR #686, opened 2026-09-16
fix: allow negative numbers in --static-position

giorgix3:
clap 4 treats values starting with a hyphen as flags unless allow_negative_numbers is set on the argument. This caused --static-position to fail for any western longitude or southern latitude, e.g.:

  --static-position 20.34 -20.12 0.0
  error: unexpected argument '-20' found

allow_negative_numbers = true is the correct fix: it permits negative numeric values while still rejecting unknown flags, unlike allow_hyphen_values which would accept any hyphen-prefixed string.

`--static-position` accepts negative latitude, longitude, and heading values while continuing to reject unknown flags.


keesverruijt:
Adding tests in followup PR

